PR #14, opened 2023-07-11
Adding Product Configuration Extension template

kmdavis:
### Background
Part of https://github.com/Shopify/core-issues/issues/57533
Related to https://github.com/Shopify/ui-extensions/pull/1069

### Solution
Adds a new template for Bundles UI Extension

### Checklist

### DIFF
--- a/product-configuration-extension/locales/en.default.json.liquid
+++ b/product-configuration-extension/locales/en.default.json.liquid
@@ -1,3 +1,3 @@
 {
-  "welcome": "Welcome to the {% raw %}{{extensionPoint}}{% endraw %} extension!"
+  "welcome": "Welcome to the {% raw %}{{target}}{% endraw %} extension!"
 }
--- a/product-configuration-extension/locales/en.default.json.liquid
+++ b/product-configuration-extension/locales/en.default.json.liquid
@@ -1,0 +1,3 @@
+{
+  "welcome": "Welcome to the {% raw %}{{extensionPoint}}{% endraw %} extension!"
+}
--- a/product-configuration-extension/locales/fr.json.liquid
+++ b/product-configuration-extension/locales/fr.json.liquid
@@ -1,3 +1,3 @@
 {
-  "welcome": "Bienvenue dans l'extension {% raw %}{{extensionPoint}}{% endraw %}!"
+  "welcome": "Bienvenue dans l'extension {% raw %}{{target}}{% endraw %}!"
 }
--- a/product-configuration-extension/locales/fr.json.liquid
+++ b/product-configuration-extension/locales/fr.json.liquid
@@ -1,0 +1,3 @@
+{
+  "welcome": "Bienvenue dans l'extension {% raw %}{{extensionPoint}}{% endraw %}!"
+}
--- a/product-configuration-extension/package.json.liquid
+++ b/product-configuration-extension/package.json.liquid
@@ -2,7 +2,6 @@
 {
   "name": "{{ handle }}",
   "version": "1.0.0",
-  "main": "dist/main.js",
   "license": "UNLICENSED",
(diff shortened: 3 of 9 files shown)


product-configuration-extension/locales/en.default.json.liquid
```diff
@@ -0,0 +1,3 @@
+{
+  "welcome": "Welcome to the {% raw %}{{extensionPoint}}{% endraw %} extension!"
```
vividviolet, line 2: As part of the Unified Config project we now need to stop saying extension points but extension targets
```suggestion
  "welcome": "Welcome to the {% raw %}{{target}}{% endraw %} extension!"
```

product-configuration-extension/locales/fr.json.liquid
```diff
@@ -0,0 +1,3 @@
+{
+  "welcome": "Bienvenue dans l'extension {% raw %}{{extensionPoint}}{% endraw %}!"
```
vividviolet, line 2:
```suggestion
  "welcome": "Bienvenue dans l'extension {% raw %}{{target}}{% endraw %}!"
```

product-configuration-extension/package.json.liquid
```diff
@@ -0,0 +1,25 @@
+{%- if flavor contains "react" -%}
+{
+  "name": "{{ name }}",
+  "version": "1.0.0",
+  "main": "dist/main.js",
```
vividviolet, line 5: We've aligned on removing this across all templates

product-configuration-extension/package.json.liquid
```diff
@@ -0,0 +1,25 @@
+{%- if flavor contains "react" -%}
+{
+  "name": "{{ name }}",
```
vividviolet, line 3: The new recommended field to use here is `handle` as it will be unique. See this other template for an example
```suggestion
  "name": "{{ handle }}",
```

product-configuration-extension/shopify.ui.extension.toml.liquid
```diff
@@ -0,0 +1,16 @@
+name = "{{ name }}"
```
vividviolet, line 1: Can you update the TOML to much the new Unified format? The file name needs to be `shopify.extension.toml.liquid` and it should follow the same format as this one: https://github.com/Shopify/extensions-templates/blob/main/block-extension/shopify.extension.toml.liquid

product-configuration-extension/shopify.ui.extension.toml.liquid
```diff
@@ -0,0 +1,16 @@
...
+type = "ui_extension"
+
+api_version = "unstable"
```
carrotderek, line 4: @vividviolet Is there a specific API version we should be using?

vividviolet (changes requested):
Unfortunately this PR is now out of date with the latest changes for Unified Config so will have to be fixed first

product-configuration-extension/shopify.ui.extension.toml.liquid
vividviolet: Yes, `2023-07` should be the target

vividviolet (approved):

product-configuration-extension/package.json.liquid
```diff
@@ -0,0 +1,24 @@
...
+    "react": "^17.0.0",
+    "@shopify/ui-extensions-react": "unstable"
+  },
+  "devDependencies": {
+    "@types/react": "^17.0.0"
+  }
+}
+{%- else -%}
+{
+  "name": "{{ name }}",
```
isaacroldan, line 16: I'm late to this review, this is wrong, it should be {{ handle }} (same as above)

PR #23, opened 2023-07-19
Updated package.json to use handle

alfonso-noriega:
### Background

The extension name property may contain invalid characters, switching to handle we ensure the package.json `name` is valid.

### Solution

Use the `handle` value as package.json `name`

### Checklist


isaacroldan:
Awesome, thank you

PR #25, opened 2023-07-19
Update action and block extension templates to use api_version 2023-07

vividviolet:


action-extension/shopify.extension.toml.liquid
```diff
@@ -1,4 +1,4 @@
-api_version = "unstable"
+api_version = "2023-07"
```
olavoasantos, line 1: We need to update the Host to serve the latest set of components to this version.

PR #28, opened 2023-07-20
Minor edits

vividviolet:
- rename action-extension to admin-action and block-extension to admin-block
- remove setting the "main" field in the package.json as it is unused


MitchLillie (approved):
🎩'ed via https://github.com/Shopify/partners/pull/49069#pullrequestreview-1539906465 LGTM!

PR #31, opened 2023-07-21
Change target names in checkout-extension template

jamesvidler:
### Background

Ensure the new `purchase.checkout.block.render` is used as opposed to `Checkout::Dynamic::Render`.

### Solution

Update the target name.

### Checklist

### DIFF
--- a/checkout-extension/shopify.extension.toml.liquid
+++ b/checkout-extension/shopify.extension.toml.liquid
@@ -29,7 +29,7 @@ block_progress = true
 
 [[extensions.targeting]]
 module = "./src/CheckoutDynamicRender.{{ srcFileExtension }}"
-target = "Checkout::Dynamic::Render"
+target = "purchase.checkout.block.render"
 
 
 
--- a/checkout-extension/src/CheckoutDynamicRender.liquid
+++ b/checkout-extension/src/CheckoutDynamicRender.liquid
@@ -7,7 +7,7 @@ import {
   Banner,
 } from '@shopify/ui-extensions-react/checkout';
 
-export default reactExtension('Cpurchase.checkout.block.render', () => <App />);
+export default reactExtension('purchase.checkout.block.render', () => <App />);
 
 function App() {
   const { extension: { target } } = useApi();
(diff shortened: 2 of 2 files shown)


checkout-extension/shopify.extension.toml.liquid
```diff
@@ -29,7 +29,7 @@ block_progress = true
 
 [[extensions.targeting]]
 module = "./src/CheckoutDynamicRender.{{ srcFileExtension }}"
```
vividviolet, line 31: Can we rename the file to PurchaseCheckoutBlockRender to match?

checkout-extension/shopify.extension.toml.liquid
```diff
@@ -1,7 +1,7 @@
 # Learn more about configuring your checkout UI extension:
 # https://shopify.dev/api/checkout-extensions/checkout/configuration
 
-api_version = "2023-07"
+api_version = "unstable"
```
vividviolet, line 4: This can stay `2023-07` because the deployment validation will still pass - it's only the package.json that needs to change

vividviolet (commented):
Just missing the updates to the package.json [here](https://github.com/Shopify/extensions-templates/blob/f42cbcb8185a332606ba9999c0af366d014261da/checkout-extension/package.json.liquid#L8) and [here](https://github.com/Shopify/extensions-templates/blob/f42cbcb8185a332606ba9999c0af366d014261da/checkout-extension/package.json.liquid#L20)

jamesvidler:
I've updated the package.json and renamed the module to **PurchaseCheckoutBlockRender**. 👍

checkout-extension/src/CheckoutDynamicRender.liquid
vividviolet, line 1: This file has to be renamed to `PurchaseCheckoutBlockRender.liquid`

jamesvidler: baaaahhh

jamesvidler: Sorry, my brain is running on fumes 😂 . Fixed.

PR #34, opened 2023-07-24
Make packages private and add ui-extensions for react templates

lemonmade:
### Background

### Solution

### Checklist

### DIFF
--- a/admin-block/package.json.liquid
+++ b/admin-block/package.json.liquid
@@ -1,10 +1,12 @@
 {%- if flavor contains "react" -%}
 {
   "name": "{{ handle }}",
+  "private": true,
   "version": "1.0.0",
   "license": "UNLICENSED",
   "dependencies": {
     "react": "^17.0.0",
+    "@shopify/ui-extensions": "unstable",
     "@shopify/ui-extensions-react": "unstable"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


admin-block/package.json.liquid
```diff
@@ -1,10 +1,12 @@
 {%- if flavor contains "react" -%}
 {
   "name": "{{ handle }}",
+  "private": true,
   "version": "1.0.0",
   "license": "UNLICENSED",
   "dependencies": {
     "react": "^17.0.0",
+    "@shopify/ui-extensions": "unstable",
```
vividviolet, line 9: Do we need the vanilla version of the library for the react template?

checkout-extension/package.json.liquid
```diff
@@ -1,10 +1,12 @@
 {%- if flavor contains "react" -%}
 {
   "name": "{{ handle }}",
+  "private": true,
   "version": "1.0.0",
   "license": "UNLICENSED",
   "dependencies": {
     "react": "^17.0.0",
+    "@shopify/ui-extensions": "unstable",
```
vividviolet, line 9: Same question here

lemonmade:
@vividviolet I think we probably do want to remove those but I will leave it to the teams that own those templates. I'd like to restrict my changes to the old templates to only this more important fix of preventing accidental publishing.

PR #35, opened 2023-07-24
Improve checkout template

lemonmade:
### Background

This PR cleans up the checkout extension template. There are a collection of unrelated changes in here, so I have written up my rationale for each as comments on the PR. The general thrust of these changes was a simpler, prettier template extension.

### Checklist

### DIFF
--- a/checkout-extension/README.md.liquid
+++ b/checkout-extension/README.md.liquid
@@ -2,87 +2,38 @@
 
 Checkout UI extensions let app developers build custom functionality that merchants can install at defined targets in the checkout flow. You can learn more about checkout UI extensions in Shopify’s [developer documentation](https://shopify.dev/api/checkout-extensions/checkout).
 
-
 ## Prerequisites
-Make sure to complete [all the prerequisites](https://shopify.dev/apps/checkout/delivery-instructions/getting-started#requirements) from any tutorials to develop a Checkout UI extension.
 
-## Tutorials
-The best way to get started is to follow some of the available Extension Tutorials:
+Before you start building your extension, make sure that you’ve created a [development store](https://shopify.dev/docs/apps/tools/development-stores) with the [checkout extensibility developer preview](https://shopify.dev/docs/api/release-notes/developer-previews#previewing-new-features).
 
-For creating an extension:
-* [Enable extended delivery instructions](https://shopify.dev/apps/checkout/delivery-instructions)
-* [Adding product offer recommendations](https://shopify.dev/apps/checkout/product-offers)
-* [Creating a custom banner](https://shopify.dev/apps/checkout/custom-banners)
+## Your new Extension
 
-To add specific features to an extension
-* [Adding field validation](https://shopify.dev/apps/checkout/validation)
-* [Localizing an extension](https://shopify.dev/apps/checkout/localize-ui-extensions)
+Your new extension contains the following files:
 
-## Getting Started
-Initially, your extension will have the following files:
+- `README.md`, the file you are reading right now.
+- `shopify.extension.toml`, the configuration file for your extension. This file defines your extension’s name, where it will appear in the checkout, and other metadata.
+- `src/Checkout.{{ srcFileExtension }}`, the source code for your extension.
+- `locales/en.default.json` and `locales/fr.json`, which contain translations used to [localized your extension](https://shopify.dev/docs/apps/checkout/best-practices/localizing-ui-extensions).
 
-```
-└── my-app
-  └── extensions
-    └── my-checkout-ui-extension
-        ├── src
-        │   └── PurchaseCheckoutBlockRender.{{ srcFileExtension }} The source code for the `purchase.checkout.block.render` extension target
-        ├── locales
-        │   ├── en.default.json // The default locale for the checkout UI extension. Shared across all extension targets.
-        │   └── fr.json // The locale file for non-regional French translations. Shared across all extension targets.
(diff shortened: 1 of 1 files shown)


checkout-extension/README.md.liquid
```diff
@@ -2,87 +2,38 @@
 
 Checkout UI extensions let app developers build custom functionality that merchants can install at defined targets in the checkout flow. You can learn more about checkout UI extensions in Shopify’s [developer documentation](https://shopify.dev/api/checkout-extensions/checkout).
 
-
 ## Prerequisites
-Make sure to complete [all the prerequisites](https://shopify.dev/apps/checkout/delivery-instructions/getting-started#requirements) from any tutorials to develop a Checkout UI extension.
```
lemonmade, line 7: The old README was quite long. I tried to narrow this down to just the most critical concepts for a developer to understand.

checkout-extension/shopify.extension.toml.liquid
```diff
@@ -1,35 +1,51 @@
...
 # https://shopify.dev/api/checkout-extensions/checkout/configuration
 
+# The version of APIs your extension will receive. Learn more:
+# https://shopify.dev/docs/api/usage/versioning
 api_version = "2023-07"
 
 [[extensions]]
-name = "{{ name }}"
 type = "ui_extension"
+name = "{{ name }}"
```
lemonmade, line 10: I find these config files look nicest with `type` as the first line, since it is generally he shortest and the only one of these three related fields that does not contain developer-specific content.

checkout-extension/shopify.extension.toml.liquid
```diff
@@ -1,35 +1,51 @@
 # Learn more about configuring your checkout UI extension:
 # https://shopify.dev/api/checkout-extensions/checkout/configuration
 
+# The version of APIs your extension will receive. Learn more:
+# https://shopify.dev/docs/api/usage/versioning
```
lemonmade, line 5: I felt it was useful to offer a link to each of the key concepts being promoted by the template config file.

checkout-extension/shopify.extension.toml.liquid
```diff
@@ -1,35 +1,51 @@
...
-# key = "my-key"
-# [[extensions.metafields]]
-# namespace = "my-namespace"
-# key = "my-other-key"
+# Controls where in Shopify your extension will be injected,
+# and the file that contains your extension’s source code. Learn more:
+# https://shopify.dev/docs/api/checkout-ui-extensions/unstable/targets-overview
+
+[[extensions.targeting]]
+module = "./src/Checkout.{{ srcFileExtension }}"
```
lemonmade, line 18: The old name of this file (`PurchaseCheckoutBlockRender.js`) was very long, and would become inaccurate if you chose any other target name. IMO, it would be nicer if we just output a file named `{{handle}}.js` to the root of the project, but this felt like an OK compromise. cc/ @vividviolet

checkout-extension/shopify.extension.toml.liquid
```diff
@@ -1,35 +1,51 @@
...
+target = "purchase.checkout.block.render"
 
 [extensions.capabilities]
+# Gives your extension access to directly query Shopify’s storefront API.
+# https://shopify.dev/docs/api/checkout-ui-extensions/unstable/configuration#api-access
 api_access = true
-network_access = true
-block_progress = true
```
lemonmade, line 21: I removed this since it's not something we want or expect every extension doing.

checkout-extension/src/Checkout.liquid
```diff
@@ -1,33 +1,38 @@
...
 } from '@shopify/ui-extensions-react/checkout';
 
-export default reactExtension('purchase.checkout.block.render', () => <App />);
+export default reactExtension(
+  'purchase.checkout.block.render',
+  () => <Extension />,
+);
 
-function App() {
-  const { extension: { target } } = useApi();
```
lemonmade, line 13: IMO, nested destructuring is a little overwhelming for folks who aren't super familiar with JavaScript (and even some that are), because it's not super common to see and doesn't always have a matching concept in other languages. I restricted myself to a single level of destructuring in these examples.

checkout-extension/src/Checkout.liquid
```diff
@@ -1,33 +1,38 @@
 {%- if flavor contains "react" -%}
-import React from 'react';
```
lemonmade, line 2: The React extensions use the `jsx-runtime` React transform, which does not require this value to be in scope. There's actually a warning about this in source code when generating an extension today.

sjanczy2 (approved):
LGTM, as mentioned in a comment I'm going to add our APIs to the https://shopify.dev/docs/api/usage/versioning page

Also had to do a double take on seeing `Checkout.liquid` as one of the edited files 😅

checkout-extension/shopify.extension.toml.liquid
lemonmade: Thank you! I had a draft item in the project for this, if you want to mark it as done: https://github.com/orgs/Shopify/projects/5920/views/1?pane=issue&itemId=33989493

checkout-extension/shopify.extension.toml.liquid
vividviolet: I'm good with this change but can you clarify how a root one with the handle name would work? I thought we needed a default export for these files so we'd need 1 per target?

vividviolet (approved):
Looks good to me

PR #53, opened 2023-09-26
Bump react version fixing error seen in prod

PrescilliaA:
### Background

- Close https://github.com/Shopify/customer-data-platform/issues/6049

When testing the Customer Segment Template extension generation in production, I encountered the following error: 
image

### Solution

Bumping the react version to 18 should fix the issue as we are matching the one in ui-extensions-react.

### Checklist

### DIFF
--- a/customer-segment-template-extension/package.json.liquid
+++ b/customer-segment-template-extension/package.json.liquid
@@ -5,12 +5,13 @@
   "version": "1.0.0",
   "license": "UNLICENSED",
   "dependencies": {
-    "react": "^17.0.0",
+    "react": "^18.0.0",
     "@shopify/ui-extensions": "unstable",
     "@shopify/ui-extensions-react": "unstable"
   },
   "devDependencies": {
-    "@types/react": "^17.0.0"
+    "@types/react": "^18.0.0",
+    "react-reconciler": "0.29.0"
   }
 }
 {%- else -%}


lauramann (approved):

PrescilliaA:
> Either way, if you depend on unstable, you should also add react-reconciler@^0.29.0 as a dependency (for react@18) or react-reconciler@^0.27.0 (if you want to keep react@17)

Thanks! I added the dependency. We will only be reliant on the unstable branch until the next version of the library publishes in october. :)

customer-segment-template-extension/package.json.liquid
```diff
@@ -20,7 +21,7 @@
   "version": "1.0.0",
   "license": "UNLICENSED",
   "dependencies": {
-    "@shopify/ui-extensions": "unstable"
+    "@shopify/ui-extensions": "unstable",
```
MitchLillie, line 24:
```suggestion
    "@shopify/ui-extensions": "unstable"
```

PrescilliaA: Removed!

MitchLillie (approved):
There's a [bigger issue](https://github.com/Shopify/ui-extensions/pull/1296#issuecomment-1731471642) going on with the `unstable` branch and react@17, but if upgrading to react@18 works then I'm happy to approve.

PR #55, opened 2023-09-27
Update ui-extension and react version.

lihaokx:
### Background

### Solution

### Checklist


customer-account-extension/locales/en.default.json
```diff
@@ -1,3 +1,3 @@
 {
-  "welcome": "Welcome to the {{target}} customer account ui extension!"
+  "welcome": "Welcome to the {{target}} in customer account ui extension!"
```
jrrafols, line 2: Will this actually be exposed to 3p devs at some point? If so, I think it could use a bit of re-wording. You may even want to ping someone on the UX team for copy.

```suggestion
  "welcome": "Welcome to the {{target}} target of your in customer account UI extension!"
```

customer-account-extension/locales/fr.json
```diff
@@ -1,3 +1,3 @@
 {
-  "welcome": "Bienvenue dans l'extension {{target}} de l'interface utilisateur du compte client!"
+  "welcome": "Bienvenue dans la {{target}} de l'extension d'interface utilisateur du compte client!"
```
jrrafols, line 2: Following from above, did this go through the translation team?

```suggestion
  "welcome": "Bienvenue à la cible {{target}} de l'extension d'interface utilisateur du compte client!"
```

jrrafols (commented):
The change itself seems fine to me, though I would defer to someone on the FE team as they have more context about whether there are any concerns with updating dependencies.

I did leave a couple comments about the actual wording, but I'm not sure if this is the template that will be exposed for Dev Preview. If we don't anticipate this version of things being seen by 3p devs, then it's less of a concern for now.

customer-account-extension/locales/en.default.json
lihaokx: > Will this actually be exposed to 3p devs at some point?

Yes. It is only a template to show, and I think 3p devs will surely change it. It is mostly copy paste from [checkout](https://github.com/Shopify/extensions-templates/blob/main/checkout-extension/locales/en.default.json).

> You may even want to ping someone on the UX team for copy.

Makes sense. But I will reach out to ux team next time when they consider multiple templates, and then change it together. Right now is not exposed to 3p, so I think it is ok.

customer-account-extension/locales/fr.json
lihaokx: >  did this go through the translation team?

No, I just use goolge translate to do it. 😄

lihaokx:
> but I'd definitely wait for a review from someone more familiar with the implications of updating the dependencies.

Sure. Will wait @robin-drexler  to have a look

customer-account-extension/locales/en.default.json
lihaokx: Done. Thanks

customer-account-extension/locales/fr.json
lihaokx: Updated.

alfonso-noriega (approved):
🎩 + LGTM

PR #59, opened 2023-10-13
[CustomerSegmentTemplate] Add title and description locales

PrescilliaA:
### Background

The `name` and `description` locales are not being automatically picked up and are not showing up in the `CustomerSegmentTemplate`.

This will better match the [tutorials](https://github.com/Shopify/shopify-dev/pull/38211) I've created for Shopify dev. 

### Solution

 I've switched the liquid file to use `i18n` to populate the `title` and `description` props.

### Checklist

### DIFF
--- a/customer-segment-template-extension/locales/en.default.json
+++ b/customer-segment-template-extension/locales/en.default.json
@@ -1,4 +1,4 @@
 {
-    "name": "First-time customers who bought specific products",
+    "title": "First-time customers who bought specific products",
     "description": "Send customers invitations to encourage them to create an account for faster checkouts, easily track order status, and view order history."
 }
--- a/customer-segment-template-extension/src/CustomerSegmentTemplate.liquid
+++ b/customer-segment-template-extension/src/CustomerSegmentTemplate.liquid
@@ -10,13 +10,17 @@ const TARGET = 'admin.customers.segmentation-templates.render';
 export default reactExtension(TARGET, () => <App />);
 
 function App() {
+  // The useApi hook provides access to several useful APIs like i18n, close, and data.
+  const { i18n } = useApi(TARGET);
+
   const query = 'number_of_orders = 1 AND products_purchased(id: (product_id)) = true';
   const queryToInsert = 'number_of_orders = 1 AND products_purchased(id: (';
 
   return (
     // The CustomerSegmentTemplate component provides an API for setting the title, description, date, query, and query to insert of the segmentation template.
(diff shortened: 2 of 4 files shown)


customer-segment-template-extension/locales/en.default.json
```diff
@@ -1,4 +1,4 @@
 {
-    "name": "First-time customers who bought specific products",
+    "title": "First-time customers who bought specific products",
```
LTiger14, line 2: Does this bring the "automated" translations?

customer-segment-template-extension/src/CustomerSegmentTemplate.liquid
```diff
@@ -10,13 +10,17 @@ const TARGET = 'admin.customers.segmentation-templates.render';
...
+
   const query = 'number_of_orders = 1 AND products_purchased(id: (product_id)) = true';
   const queryToInsert = 'number_of_orders = 1 AND products_purchased(id: (';
 
   return (
     // The CustomerSegmentTemplate component provides an API for setting the title, description, date, query, and query to insert of the segmentation template.
-    // The title and description do not need to be explicitly set as they will be obtained from the extension's configuration file (TOML).
     <CustomerSegmentTemplate
+        title={i18n.translate("title")}
+        description={i18n.translate("description)}
```
LTiger14, line 23:
```suggestion
        description={i18n.translate("description")}
```

LTiger14: One more note on single vs double quote. Nothing breaking it's just the date uses a single one.

customer-segment-template-extension/locales/en.default.json
LTiger14, line 2: About this rename https://github.com/Shopify/extensions-templates/blob/main/customer-segment-template-extension/shopify.extension.toml.liquid#L5 is still `t:name`. I'm wondering if it makes sense to use the translation there. I know @vividviolet talked about something there and the `description` support being missing. I'm wondering if something evolved there.

I'm thinking it looks weird if to use this as an app title if we encourage multiple templates per `ts` file. 🤔

lauramann (approved):
LGTM :)

customer-segment-template-extension/locales/en.default.json
PrescilliaA: I agree, the t:name comes from the CLI command when naming the extension. I changed it up so now the extension name is independent of the template title. I thought this was a better solution, at least until there is proper support for the description field.

alfonso-noriega (approved):
LGTM!

PR #61, opened 2023-10-26
Update customer account template

lihaokx:
### Background

Update customer account ui extension template

### Tophat

Prerequisite: enable customer account flag of you partner account in the partner internal page.

1. In your local extension fold, run the following command:

`npm run shopify app generate extension -- --template customer_account_ui  --clone-url https://github.com/Shopify/extensions-templates#update-customer-account-template`

2. Then, choose your flavor of extension. Check if the 7 targets work well on your dev store.

image

### Solution

Add 7 targets to customer account template.

### Checklist


customer-account-extension/shopify.extension.toml.liquid
```diff
@@ -6,5 +6,29 @@ name = "{{ name }}"
 handle = "{{ handle }}"
```
kirakik, line 7: Don't we need to update the `api_version` at the top of this file to be "unstable"?

kirakik: Also, checkout includes the `api_access` capability [by default](https://github.com/Shopify/extensions-templates/blob/main/checkout-extension/shopify.extension.toml.liquid#L24), we should probably do the same. And [add commented out config](https://github.com/Shopify/extensions-templates/blob/main/checkout-extension/shopify.extension.toml.liquid#L29) for `network_access` with a link to our docs

customer-account-extension/src/OrderActionMenuItem.liquid
```diff
@@ -0,0 +1,28 @@
...
+  import { Button, reactExtension, useApi } from '@shopify/ui-extensions-react/customer-account';
+
+  export default reactExtension(
+    "customer-account.order.action.menu-item.render",
+    () => <OrderActionMenuItem />
+  );
+  
+  function OrderActionMenuItem() {
+    const {i18n} = useApi()
+    return <Button to='/'>{i18n.translate("menuItemButton")}</Button>;
```
kirakik, line 11: I think we wanted this order action menu to redirect to the full page extension

lihaokx: > I think we wanted this order action menu to redirect to the full page extension

Yes, When setting `to='/'`, it will be redirected to the full page extension after clicking it. @kirakik

kirakik: That is... very unintuitive. Is there a clearer way to do this we could use in the default example here?

customer-account-extension/locales/en.default.json
```diff
@@ -1,3 +1,4 @@
 {
-  "welcome": "Welcome to the {{target}} target of your customer account UI extension!"
+  "welcome": "Welcome to {{target}} target of your customer account UI extension ^_^",
+  "menuItemButton": "Open full page extension"
```
kirakik, line 3: Can we align with the UX guidelines here and keep this copy short so that it fits in one line? Both this and the French copy make the default thing partners see look pretty bad. 

image

customer-account-extension/src/OrderActionMenuItem.liquid
brianshen1990: we can replace it with `extension:/` to make it more clear that we want to go to the root path of the full page extension

henrytao-me (approved):
No objection from app-ui team

kirakik:
Not sure what I'm doing wrong, but a few targets are not populating correctly for me:

- customer-account.order-index.block.render
- customer-account.profile.block.render
- customer-account.profile.addresses.render-after

Are they displaying for you?

lihaokx:
It works well on my side.
> customer-account.order-index.block.render
> customer-account.profile.block.render

These two targets are dynamic. Do you set the [placement-reference](https://github.com/Shopify/customer-account-web/blob/main/app/foundation/extensibility/types.ts#L290-L299)? like; `?placement-reference=PAGE_TITLE`. @kirakik 

> customer-account.profile.addresses.render-after

This one is shown on the profile page under the addresses card.  
image

kirakik:
> It works well on my side.
> 
> > customer-account.order-index.block.render
> > customer-account.profile.block.render
> 
> These two targets are dynamic. Do you set the [placement-reference](https://github.com/Shopify/customer-account-web/blob/main/app/foundation/extensibility/types.ts#L290-L299)? like; `?placement-reference=PAGE_TITLE`. @kirakik
> 
> > customer-account.profile.addresses.render-after
> 
> This one is shown on the profile page under the addresses card. image

Weird 🤔 yeah I did use the placement reference query param, but still nothing is showing up

customer-account-extension/src/Page.liquid
```diff
@@ -0,0 +1,28 @@
+{%- if flavor contains "react" -%}
+  import { Banner, reactExtension, useApi } from '@shopify/ui-extensions-react/customer-account';
```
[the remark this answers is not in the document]
lihaokx: Done

Issue #66, opened 2023-11-02
Improve POS UI Extension Template


NathanJolly:
Extension question for the team. <https://apps.shopify.com/return-without-receipt?search_id=b0619ddf-5916-40a8-93a9-38f7899573a8&amp;surface_detail=riess&amp;surface_inter_position=1&amp;surface_intra_position=1&amp;surface_type=search|I saw this new app in WxM> and love to see more POS UI extensions in the wild :raised_hands: However, I still don't fully understand how <https://screenshot.click/02-53-ltvqg-17kn0.png|some of this alignment> is possible? This could be an opportunity to see if we're missing something even in our documentation around inset recommendations, etc? Or is there something we could do at the component level?

Anyway, not looking to diminish an overall win in new extensions being launched :rocket_animated:

[Slack Message](https://shopify.slack.com/archives/C04H9AV0KV5/p1698947800551039?thread_ts=1698947800.551039&cid=C04H9AV0KV5)

PR #68, opened 2023-11-08
Pin pos version to 1.5.0 for new extensions, and update templates to …

NathanJolly:
…include ScrollView

### Background

The current version pinning allows users to unknowingly update to 1.5.1, which is a problem version right now.

Also, while I was in there, I figured I should wrap the Text in a ScrollView, per:
https://github.com/Shopify/extensions-templates/issues/66 

### Solution

We should pin new extensions to 1.5.0 to prevent some amount of damage.
Prettier also did a bit of editing, but I did verify that both templates work as expected.

### Checklist

- [X] I have squashed my commits into chunks of work with meaningful commit messages

### DIFF
--- a/pos-ui-extension/package.json.liquid
+++ b/pos-ui-extension/package.json.liquid
@@ -7,7 +7,7 @@
   "license": "UNLICENSED",
   "dependencies": {
     "react": "^17.0.0",
-    "@shopify/retail-ui-extensions-react": "^1.4.0"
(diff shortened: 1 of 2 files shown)


vctrchu (approved):

alfonso-noriega (commented):
LGTM

alfonso-noriega (approved):

PR #73, opened 2023-12-07
Update product-configuration template

siddhantbajaj:
### Background

### Solution

Before:

Screenshot 2023-12-07 at 9 36 09 AM

After:
Screenshot 2023-12-07 at 9 33 22 AM

### Checklist

### DIFF
--- a/product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
+++ b/product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
@@ -1,4 +1,5 @@
 {%- if flavor contains "react" -%}
+import React = require('react');
 import {
   useApi,
   Text,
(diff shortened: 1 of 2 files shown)


product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
```diff
@@ -1,39 +1,84 @@
...
-export default reactExtension('admin.product-details.configuration.render', () => <App />);
-{% endif %}
+render('admin.product-details.configuration.render', () => <App />);
+
+// The main App component
 function App() {
+  // Uncomment the following line if you need to access the API
   {% if flavor contains "typescript" %}
-  const {extension: {target}, i18n} = useApi<'admin.product-details.configuration.render'>();
+  // const {extension: {target}, i18n} = useApi<'admin.product-details.configuration.render'>();
```
carrotderek, line 23: Do we need to comment this out?

product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
```diff
@@ -1,39 +1,84 @@
...
+            <BlockStack gap="none" padding="none">
+              <Text fontWeight="bold-200">
+                {component}
+              </Text>
+            </BlockStack>
+          </Box>
+        </InlineStack>,
+        index < bundleComponents.length - 1 && <Divider key={`divider-${index}`} />
+      ])}
+    </BlockStack>
```
carrotderek, line 46: Can we use `<ResourceList />` and `<ResourceItem />` components? It would simplify the markup here and I think they are available in the `ui-extensions-react` library.

carrotderek (commented):
Great start to this. Mainly question around whether we can simplify the markup by using the `ResourceList` to render the bundle components.

product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
siddhantbajaj: Yes, I missed commenting that out.

product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
siddhantbajaj: Doesn't look like it exists there - https://shopify.slack.com/archives/C04K7BZDH3N/p1702483564729509?thread_ts=1702483485.885719&cid=C04K7BZDH3N

product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
```diff
@@ -1,39 +1,84 @@
...
+          <Box padding="large none">
+            <BlockStack gap="none" padding="none">
+              <Text fontWeight="bold-200">
+                {component}
+              </Text>
+            </BlockStack>
+          </Box>
+        </InlineStack>,
+        index < bundleComponents.length - 1 && <Divider key={`divider-${index}`} />
+      ])}
```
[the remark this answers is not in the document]
siddhantbajaj: Updated. I also had to add an import for react:

```
import React = require('react');
```

Let me know If I'm doing something wrong.

product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
```diff
@@ -1,39 +1,87 @@
 {%- if flavor contains "react" -%}
+import React = require('react');
```
carrotderek, line 2: ES modules is the standard when working within the browser.

```suggestion
import React from 'react';
```

carrotderek (approved):
Looks like we're going to have to without `ResourceList` then 😆  - thanks for tracking that down. 

Just need to fix the import statement for React and we should be good.

kmdavis:
Replaced by #206

PR #79, opened 2024-01-03
Create payments app extension credit card template

marklevi:
### Background

Related issue: https://github.com/Shopify/payments-platform/issues/4378
Related project: https://docs.google.com/document/d/1Y2Ly5sw6MalZQP8w3yraQlOs8dK3UK38LVQvQz7fC_M/edit#heading=h.8coln95uazox
Related management class: https://github.com/Shopify/shopify/blob/main/components/payment_processing/payments_partners/app/models/payments_partners/payments_app_extensions/credit_card_payments_app_extension.rb

Creates an credit card payments app extension template to be used via the CLI

### Solution

### Checklist

### DIFF
--- a/payments-app-extension-credit-card/shopify.extension.toml.liquid
+++ b/payments-app-extension-credit-card/shopify.extension.toml.liquid
@@ -1,0 +1,24 @@
+# The version of APIs your extension will receive. Learn more:
+# https://shopify.dev/docs/api/usage/versioning
+api_version = "2021-07"
+
+[[extensions]]
+name = "{{ name }}"
+type = "payments_extension"
+handle = "{{ handle }}"
+
+[[extensions.targeting]]
+target = "payments.credit-card.render"
+
+merchant_label = "Credit Credit Payments App Extension"
+payment_session_url = "https://api.paymentprovider.com/endpoint"
+refund_session_url = "https://api.paymentprovider.com/refund"
+capture_session_url = "https://api.paymentprovider.com/capture"
+void_session_url = "https://api.paymentprovider.com/void"
+supported_countries = ["US"]
+supported_payment_methods = ["visa"]
+supports_3ds = false
+supports_installments = false
+supports_deferred_payments = false
+test_mode_available = true
+encryption_certificate = {}


payments-app-extension-credit-card/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,24 @@
...
+start_payment_session_url = "https://api.paymentprovider.com/endpoint"
+start_refund_session_url = "https://api.paymentprovider.com/refund"
+start_capture_session_url = "https://api.paymentprovider.com/capture"
+start_void_session_url = "https://api.paymentprovider.com/void"
+start_verification_session_url = "https://api.paymentprovider.com/verification"
+supported_countries = ["US"]
+supported_payment_methods = ["visa"]
+supports_3ds = false
+supports_installments = false
+supports_deferred_payments = false
```
wikiti, line 17: Seems like the UI groups these two together. Is there a problem with that?

payments-app-extension-credit-card/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,24 @@
...
+test_mode_available = true
+api_version = "2021-07"
+encryption_certificate = {}
+multiple_capture = false
+confirmation_callback_url = "https://api.paymentprovider.com/confirm"
+ui_extension_registration_uuid = "6f5af7b2-0946-4873-a8a8-089f66414297"
```
wikiti, line 23: What's this registration uuid? The UI extension?

payments-app-extension-credit-card/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,24 @@
...
+start_void_session_url = "https://api.paymentprovider.com/void"
+start_verification_session_url = "https://api.paymentprovider.com/verification"
+supported_countries = ["US"]
+supported_payment_methods = ["visa"]
+supports_3ds = false
+supports_installments = false
+supports_deferred_payments = false
+test_mode_available = true
+api_version = "2021-07"
+encryption_certificate = {}
```
wikiti, line 20: What information does this hash include? It would be nice to provide an example of its format in this template, since this field will be required for credit card payment apps

wikiti: I was expecting a gid or uuid, since the partner simply selects an existing certificate within the dashboard 🤔 Could this be a security issue? Do we validate in the backend if the certificate in the configuration belongs to the partner?

payments-app-extension-credit-card/shopify.extension.toml.liquid
marklevi: yes, good point - i think we might need to introduce validation in CLI or partners for instance

payments-app-extension-credit-card/shopify.extension.toml.liquid
marklevi: it maps to the registration uuid

payments-app-extension-credit-card/shopify.extension.toml.liquid
marklevi: as far as i can tell there is a light form of validation (mostly on the presence of keys and that the certificate string starts with certain characters). the frontend in this case loads partner certificates and offers the user to pick the appropriate cert.

payments-app-extension-credit-card/shopify.extension.toml.liquid
isaacroldan: it shouldn't have a default value then no?

marklevi: i think i'll remove it for now. we have a ticket to deal with this: https://github.com/Shopify/payments-platform/issues/4383

payments-app-extension-credit-card/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,28 @@
+# The version of APIs your extension will receive. Learn more:
+# https://shopify.dev/docs/api/usage/versioning
+api_version = "2021-07"
+
+[[extensions]]
+name = "{{ name }}"
+type = "payments_app_extension"
```
raluka, line 7: I think the specification linked in the PR description is of type `payments_app_credit_card`

payments-app-extension-credit-card/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,28 @@
...
+handle = "{{ handle }}"
+
+[[extensions.targeting]]
+target = "credit_card"
```
raluka, line 11: Not sure babout this syntax, looks like the target is missing the initial part (what is targeted)

Screenshot 2024-01-10 at 14 40 32

You can find some examples here: https://github.com/Shopify/extensions-templates/blob/main/checkout-extension/shopify.extension.toml.liquid

payments-app-extension-credit-card/shopify.extension.toml.liquid
marklevi: the believe the type should correspond to the specification identifier: https://github.com/Shopify/shopify/pull/469870/files#diff-348901488c164aca2dace245d34a0ad18c26750b66693a10b71c5012974dc17fR6

raluka: Oh, I see! From the PR description I thought the management class is `PaymentsPartners::PaymentsAppExtensions::CreditCardPaymentsAppExtension`, https://github.com/Shopify/shopify/blob/main/components/payment_processing/payments_partners/app/models/payments_partners/payments_app_extensions/credit_card_payments_app_extension.rb, which I found it connected only with  ["Payments App - Credit Card"](https://github.com/Shopify/shopify/blob/470f17bf4a59cfc8bdd4e49126b19c2408a5ba59/components/payment_processing/payments_partners/config/initializers/app_extensions.rb#L37) specification.

payments-app-extension-credit-card/shopify.extension.toml.liquid
bazay: Hey @raluka, thanks for sharing. I believe the concept of "target" for a payments extension differs from that of a UI or Function target, in that the "target" represents the specific type of payment extension. In moving the payment extension to the CLI we are replacing the 5 dashboard configurations in core with a single CLI configuration in core. Therefore for payment extensions I see something like:
```
a. What's targeted          |           b. Action
payment type                |           shared payment verbs + specific verbs depending on payment type

e.g.
1. offsite
2. custom onsite
3. creditcard
4. custom creditcard
5. redeemable
```

marklevi: Also, the problem with using the 
`[[extensions.targeting]]` 
convention is that it requires a [module parameter](https://shopify.dev/docs/apps/app-extensions/configuration#targets) which does not apply to us

isaacroldan (approved):

PR #56, opened 2023-10-02
Update checkout-ui extension template for version 2023-10

jamesvidler:
### Background

Updates the extension template used for checkout-ui-extensions. As of `2023-10`, extensions now use React 18.

### Solution

Update the templates used by the CLI.

### Checklist

- [ ] I have :tophat:'d these changes
- [ ] I have squashed my commits into chunks of work with meaningful commit messages

PR #85, opened 2024-01-05
Update checkout UI extension version to 2024-01

jamesvidler:
### Background

Updates the checkout ui extension template to use the new 2024-01 API version. 

### Solution

Similar to this [previous PR](https://github.com/Shopify/extensions-templates/pull/56) that did the same for version 2023-10.

This was tophatted by scaffolding a new extension and making these same changes in there, running it locally and also installing it on a shop.

Screenshot 2024-01-05 at 1 58 40 PM

### Checklist

- [X] I have squashed my commits into chunks of work with meaningful commit messages

### DIFF
--- a/checkout-extension/package.json.liquid
+++ b/checkout-extension/package.json.liquid
@@ -6,8 +6,8 @@
   "license": "UNLICENSED",
   "dependencies": {
     "react": "^18.0.0",
-    "@shopify/ui-extensions": "2023.10.x",
-    "@shopify/ui-extensions-react": "2023.10.x"
+    "@shopify/ui-extensions": "2024.1.x",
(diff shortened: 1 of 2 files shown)


lemonmade (approved):
LGTM, do you know if there are any other UI extension templates we should be asking other teams to update/ updating at the same time?

jamesvidler:
> LGTM, do you know if there are any other UI extension templates we should be asking other teams to update/ updating at the same time?

@lemonmade - Yes, there is already [another PR for admin UI extensions](https://github.com/Shopify/extensions-templates/pull/85#pullrequestreview-1806781644). Will double-check with customer-accounts, POS is still on unstable only.

PR #90, opened 2024-02-15
use TARGET directly, log data, and update deprecated methods

elanalynn:
### Background

This PR uses the `TARGET` constant in the body of the extension directly (rather than getting the string from the API), updates some deprecated methods, and logs data in the action extension to be more consistent with the block example.   

### Solution

- Use `TARGET` string directly in rendered jsx
- Update deprecated methods: `appendChild` -> `append` and `updateText` -> `update`
- Add `console.log` to log data in action extension

### Checklist


MitchLillie (approved):
LGTM!

PR #93, opened 2024-03-13
Editor extension collection template

oluwatimio:
### Background

Create a template for editor extension collection. I need this PR shipped to continue work on the partner's side so we can generate this extension

🎩
There is no way to tophat at this moment since we need the template to generate the extension.

This template will be behind a beta flag and will be used in partners. 

### Checklist

### DIFF
--- a/editor-extension-collection/shopify.extension.toml.liquid
+++ b/editor-extension-collection/shopify.extension.toml.liquid
@@ -5,6 +5,10 @@ handle = "{{ handle }}"
 # Add the handles of the extensions you want in this collection here
 in_collection=[]
 
+# You can also add extensions to the collection this way
+#[[extensions.in_collection]]
+#handle=""
+
 # A second editor extension collection added in the same extension toml config
 # [[extensions]]
 # name = "t:collection_two_name"


editor-extension-collection/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,6 @@
+[[extensions]]
```
nickwesselman, line 1: I wonder if we should include a second collection in here commented out as well, to demonstrate that they can copy/paste to add more, instead of using the CLI again?

(We should obviously verify this works as expected too.)

editor-extension-collection/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,6 @@
...
+name = "{{ name }}"
```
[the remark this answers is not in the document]
nickwesselman: Yeah we do it in Functions ([example](https://github.com/Shopify/function-examples/blob/main/discounts/javascript/product-discounts/default/shopify.extension.toml.liquid#L5)) and I believe UI extensions as well.

editor-extension-collection/shopify.extension.toml.liquid
nickwesselman: hmm nested arrays in TOML can get messy, but adding another `[[extensions]]` element to the same file should be OK. The CLI _should_ just pick it up as an additional extension.

We can see how this looks when implemented though!

oluwatimio: Sounds good, just to clarify so for now will you like me to add a comment for a second extension or we can leave it out?

nickwesselman: Why don't we put one in as a reminder to test it out, it can always be removed later.

oluwatimio: Updated the template! Feel free to check it out here https://github.com/Shopify/extensions-templates/pull/93/commits/7ce5534a71299c0140c937a14d0ebe197a7d7467. Also updated the locale so it has two collection name keys

editor-extension-collection/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,13 @@
+[[extensions]]
+name = "t:collection_one_name"
+type = "editor_extension_collection"
```
Johnkinging, line 3: I think we should exclude `editor` from the type name. Possibly even `module` to `extension`, if we really trying to generically template this. 

the use case of course is specifically extensions in the editor for now, but as a module type this is more versatile going forward.

oluwatimio: Hey John 👋🏽 thanks for the feedback. So the typename here is the name we all aligned on in the tech doc. It was reviewed by @nickwesselman @deannatron and shared with the build ux team. You can view the reasoning behind it in that doc. That being said, we agreed that editor naming was sufficient

deannatron: The recommendation was that the name be inclusive of themes and POS as well, both of which have editors too, in case they want to leverage the same mechanism for surfacing collections to merchants in the editor, with the same pattern, so the recommendation was not to be specific to the checkout and accounts editor.

Johnkinging: Makes sense. Casts a decent sized net. I still don't think we need to even specify anything to with editors, but there's at least good reasons.

thanks for the context both!

editor-extension-collection/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,6 @@
+[[extensions]]
+name = "{{ name }}"
+type = "editor_extension_collection"
+handle = "{{ handle }}"
+# Add the handles of the extensions you want in this collection here
+in_collection=[]
```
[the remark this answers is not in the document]
Johnkinging: What about a commented out example?

Johnkinging (approved):

alfonso-noriega (approved):
LGTM

editor-extension-collection/shopify.extension.toml.liquid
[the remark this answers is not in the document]
oluwatimio: Makes sense, will do that here as well

PR #97, opened 2024-03-26
add new action extension targets to template comment

JoshuaWhite1:
Adding the new extension targets to the toml file of of the action extension template.

image

### Tophat instructions
- create an app
- generate an extension via my branch 
- `npm run generate extension -- --clone-url https://github.com/Shopify/extensions-templates\#jw/new-action-targets`
- inspect the toml file for new targets

### Checklist

### DIFF
--- a/admin-action/shopify.extension.toml.liquid
+++ b/admin-action/shopify.extension.toml.liquid
@@ -11,7 +11,10 @@ module = "./src/ActionExtension.{{ srcFileExtension }}"
 target = "admin.product-details.action.render"
 
 
-# Valid extension points:
+# Valid extension targets:
+
+# Abandoned checkout details page
+# - admin.abandoned-checkout-details.action.render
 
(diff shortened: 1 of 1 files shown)


admin-action/shopify.extension.toml.liquid
```diff
@@ -11,20 +11,40 @@ module = "./src/ActionExtension.{{ srcFileExtension }}"
...
+
+# Collection index and detail pages
+# - admin.collection-details.action.render
+# - admin.collection-index.action.render
+
+# Order index, detail pages and order fulfilled card
 # - admin.order-index.action.render
 # - admin.order-index.selection-action.render
 # - admin.order-details.action.render
+# - admin.order-fulfilled-card.action.render
```
JoshuaWhite1, line 42: wasn't sure if this should be in it's own category since it's kind of its own thing despite being located so closely to the others

PR #109, opened 2024-05-29
Payment app TOML file enhancements

bazay:
### Background

Changes include:
- Use latest api_version
- Add UI Extension commented code in onsite & custom CC

### Solution

See description

### Checklist

### DIFF
--- a/payments-app-extension-redeemable/shopify.extension.toml.liquid
+++ b/payments-app-extension-redeemable/shopify.extension.toml.liquid
@@ -1,6 +1,6 @@
 # The version of APIs your extension will receive. Learn more:
 # https://shopify.dev/docs/api/usage/versioning
-api_version = "2023-07"
+api_version = "2024-04"
 
 [[extensions]]
 name = "{{ name }}"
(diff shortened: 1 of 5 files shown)


payments-app-extension-redeemable/shopify.extension.toml.liquid
```diff
@@ -1,6 +1,6 @@
 # The version of APIs your extension will receive. Learn more:
 # https://shopify.dev/docs/api/usage/versioning
-api_version = "2023-07"
+api_version = "2024-04"
```
marklevi, line 4: perhaps add `checkout_payment_method_fields` in redeemables as well as it is also a part of the configuration: https://github.com/Shopify/shopify/blob/main/components/payment_processing/payments_partners/app/models/payments_partners/payments_app_extensions/redeemable_payments_app_extension.rb#L51

bazay: Added 👍

marklevi (approved):

PR #112, opened 2024-06-05
Update print action template

elanalynn:
### Background

Gets print extension template into a more final state. There will still be some work to remove the name. 

### Solution

Match designs.

### Checklist


MitchLillie (approved):
🎩 LGTM!

PR #113, opened 2024-06-06
Templates JSON

amcaplan:
### Background

For the App Management API, we are moving to a system with template specifications in a static JSON file + client-side enforcement of feature gating. More details [here](https://github.com/Shopify/cli/pull/3923)

### Solution

Created a static JSON file with the templates. This will currently live side-by-side with Partners but will eventually replace it.

### Checklist

### DIFF
--- a/templates.json
+++ b/templates.json
@@ -380,5 +380,237 @@
         ]
       }
     ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


amcaplan:
Yes, I copied that order and verified it works correctly!

PR #115, opened 2024-06-10
Update print extension template

elanalynn:
### Background

Finalizing print action extension template. 

### Solution

- Uses unstable
- Adds comments

### Checklist


admin-print-action/src/PrintActionExtension.liquid
```diff
@@ -24,9 +24,19 @@ function App() {
...
 
+  /* 
+    This template fetches static documents from the CDN to demonstrate printing.
+    However, when building your extension, you should update the src document
+    to match the resource that the user is printing. You can do this by getting the 
+    resource id from the data API and using it to create a URL with a path to your app 
+    that shows the correct document. For example, you might use a URL parameter to 
+    render an invoice for a specific order.
+    
+    https://www.example.org/print/invoice&orderId={data.selected}
```
MitchLillie, line 38:
```suggestion
    https://www.example.org/print/invoice&orderId={data.selected[0].id}
```

or 
```
    https://www.example.org/print/invoice&orderIds={data.selected.join(',')}
```

admin-print-action/src/PrintActionExtension.liquid
```diff
@@ -158,6 +175,16 @@ extend(TARGET, (root, { i18n, data }) => {
...
```
MitchLillie, line 186: Same change as above

Also, since we support relative URLs, we could also just leave off the example.org bit.

MitchLillie (commented):
A couple small tweaks 🙏

AaronSpringut (approved):

PR #119, opened 2024-06-17
Remove name translateions and add comment in toml

elanalynn:
### Background

We do not use `name` in the print extension but we can't remove it from the `toml` because we use it to validate ui-extensions.  

### Solution

This PR adds a comment in the `toml` and removes the translations. 

### Checklist

### DIFF
--- a/admin-print-action/shopify.extension.toml.liquid
+++ b/admin-print-action/shopify.extension.toml.liquid
@@ -1,12 +1,12 @@
 api_version = "2024-04"
 
 [[extensions]]
-name = "{{ handle }}"
+name = ""
 handle = "{{ handle }}"
 type = "ui_extension"
 {% if uid %}uid = "{{ uid }}"{% endif %}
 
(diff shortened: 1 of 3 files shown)


admin-print-action/shopify.extension.toml.liquid
```diff
@@ -1,12 +1,13 @@
 api_version = "2024-04"
 
 [[extensions]]
+# Name is not used in admin print actions, but cannot be removed because of validation requirements.
```
MitchLillie, line 4: Do we care to even mention this?

MitchLillie (approved):
One comment, not blocking, LGTM!

admin-print-action/shopify.extension.toml.liquid
elanalynn: I thought it might be useful if someone was trying to track down where it's used. I'm happy to remove it too.

PR #123, opened 2024-07-05
Update admin templates to use 2024-07 release

MitchLillie:
### Background

Relates to https://github.com/Shopify/app-ui/issues/1340

### Solution

### Checklist

### DIFF
--- a/admin-action/package.json.liquid
+++ b/admin-action/package.json.liquid
@@ -6,8 +6,8 @@
   "license": "UNLICENSED",
   "dependencies": {
     "react": "^18.0.0",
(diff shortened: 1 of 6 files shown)


alfonso-noriega (approved):

vividviolet (approved):
Looks good. Can you also update the minimum [here](https://github.com/Shopify/shopify/blob/0e01c7e6ea59d0c32cccd903a4b68bf362f0433a/components/apps/app/services/apps/models/ui_extension/schema/groups/print_action_extension.rb#L24)? It should be `2024-07`

PR #124, opened 2024-07-08
Update checkout UI extension template (2024-07)

jamesvidler:
### Background

Part of https://github.com/Shopify/checkout-web/issues/35555

Now that [2024-07 has been released](https://shopify.dev/docs/api/checkout-ui-extensions/2024-07/instructions-update). It is important that 3p devs check for [cart instructions](https://shopify.dev/docs/api/checkout-ui-extensions/2024-07/apis/cart-instructions) before using certain APIs. Failure to do so could mean they are rendering UI extensions in certain contexts where they won't work as expected.

### Solution

This updates the template used to generate the code for checkout UI extensions. Instead of simply rendering a banner as it did before, this will now render that same banner in addition to a "checkbox" to request a free gift (saved using checkout attributes). This uses **cart instructions** to check whether attributes are available. If they are not, it renders a warning banner for demonstration purposes.

**Before:**
Screenshot 2024-07-08 at 12 51 52 PM

**After (`canUpdateCartAttributes===true`):**
Screenshot 2024-07-08 at 12 48 50 PM

**After (`canUpdateCartAttributes===false`):**
Screenshot 2024-07-08 at 12 49 23 PM

### Checklist

### DIFF
--- a/checkout-extension/src/Checkout.liquid
+++ b/checkout-extension/src/Checkout.liquid
@@ -1,39 +1,126 @@
 {%- if flavor contains "react" -%}
 import {
-  Banner,
-  useApi,
-  useTranslate,
   reactExtension,
-} from '@shopify/ui-extensions-react/checkout';
+  Banner,
+  BlockStack,
+  Checkbox,
+  Text,
+  useApi,
+  useApplyAttributeChange,
+  useInstructions,
+  useTranslate,
+} from "@shopify/ui-extensions-react/checkout";
 
-export default reactExtension(
-  'purchase.checkout.block.render',
-  () => <Extension />,
-);
+// 1. Choose an extension target
+export default reactExtension("purchase.checkout.block.render", () => (
+  <Extension />
+));
 
 function Extension() {
   const translate = useTranslate();
   const { extension } = useApi();
+  const instructions = useInstructions();
+  const applyAttributeChange = useApplyAttributeChange();
 
+
+  // 2. Check instructions for feature availability, see https://shopify.dev/docs/api/checkout-ui-extensions/apis/cart-instructions for details
+  if (!instructions.attributes.canUpdateAttributes) {
+    // For checkouts such as draft order invoices, cart attributes may not be allowed
+    // Consider rendering a fallback UI or nothing at all, if the feature is unavailable
+    return (
+      <Banner title="{{ name }}" status="warning">
+        {translate("attributeChangesAreNotSupported")}
+      </Banner>
+    );
+  }
+
+  // 3. Render a UI
   return (
-    <Banner title="{{ name }}">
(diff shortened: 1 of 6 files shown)


checkout-extension/src/Checkout.liquid
```diff
@@ -1,39 +1,124 @@
...
+  const instructions = useInstructions();
+  const applyAttributeChange = useApplyAttributeChange();
+
 
+  // 2. Check instructions for feature availability
+  if (!instructions.attributes.canUpdateAttributes) {
+    // Consider rendering a fallback UI or nothing at all, if the feature is unavailable
+    return (
+      <Banner title="{{ name }}" status="warning">
+        Attribute changes are not supported in this checkout
```
kumar303, line 31: This could use `translate()`

checkout-extension/src/Checkout.liquid
```diff
@@ -1,39 +1,124 @@
...
-      {translate('welcome', {target: extension.target})}
-    </Banner>
+    <BlockStack border={"dotted"} padding={"tight"}>
+      <Banner title="{{ name }}">
+        {translate("welcome", {
+          target: <Text emphasis="italic">{extension.target}</Text>,
+        })}
+      </Banner>
+      <Checkbox onChange={onCheckboxChange}>
+        {translate("requestGift")}
```
kumar303, line 45: nit: I like when translation keys are fully readable:

```suggestion
        {translate("iWouldLikeAFreeGiftWithMyOrder")}
```

checkout-extension/src/Checkout.liquid
```diff
@@ -1,39 +1,124 @@
...
-  );
+  // 2. Check instructions for feature availability
+  api.instructions.subscribe((instructions) => {
+    if (!instructions.attributes.canUpdateAttributes) {
+      // Consider rendering a fallback UI or nothing at all, if the feature is unavailable
+      root.replaceChildren(
+        root.createComponent(
+          Banner,
+          { title: "{{ name }}", status: "warning" },
+          "Attribute changes are not supported in this checkout"
```
kumar303, line 80: Same comment about `translate()`

kumar303 (approved):
Looks good ✨ just some minor comments

checkout-extension/src/Checkout.liquid
jamesvidler: It could. My thinking was that you'd probably never actually output this to the buyer though and is more for demonstration purposes within the extension.

kumar303: I just thought that since it's a banner then it would add consistency. But, yeah, no biggie.

mleandres (approved):
nice!

PR #129, opened 2024-07-26
Update customer account template to 1 target

lihaokx:
### Background

### Solution

### Checklist


customer-account-extension/src/OrderStatusBlock.liquid
```diff
@@ -0,0 +1,67 @@
...
+
+function PromotionBanner() {
+  const { i18n } = useApi();
+
+  return (
+    <Banner>
+      <BlockStack inlineAlignment="center" >
+        <TextBlock>
+          {i18n.translate("earnPoints")}
+          <Link>
```
robin-drexler, line 24: A link without a target is weird, isn't it?

Maybe we remove the `view rewards` part?

customer-account-extension/src/OrderStatusBlock.liquid
```diff
@@ -0,0 +1,67 @@
...
+    {inlineAlignment: "center"}
+  )
+
+  const text = root.createComponent(
+    TextBlock,
+    {},
+    i18n.translate("earnPoints")
+  )
+
+  const link = root.createComponent(
```
robin-drexler, line 56: If we decide to remove the link, we need to do it here too.

customer-account-extension/src/OrderStatusBlock.liquid
vividviolet: Where is this meant to go? Might be a good place to showcase our custom protocols like `app`, `shopify` or `extension` since not many partners know about them

robin-drexler: We don't have `app:`, `extension:` doesn't really make sense because we're only adding a single extension point. 😅 
`shopify:` we could do. We could link to the order list page, but with what text?

maybe @deannatron has a content idea that would give as a good excuse to put a link to host page like order list or the profile page.

lihaokx: > A link without a target is weird, isn't it?
> 
> Maybe we remove the view rewards part?

Good point, this is a copy from our [public inline-extension doc](https://shopify.dev/docs/apps/build/customer-accounts/inline-extensions/build-order-status?extension=react#preview-the-extension), whose link does not have a place redirect to. We may need to update the doc as well.

customer-account-extension/src/OrderStatusBlock.liquid
lihaokx: Link is removed.

brianshen1990 (approved):

PR #130, opened 2024-08-01
Add marketing activity extension template

reemrazak:
### Background
Screenshot 2024-08-06 at 4 48 28 PM

Add marketing activity extension template

### Solution

### Checklist

### DIFF
--- a/marketing-activity/shopify.extension.toml.liquid
+++ b/marketing-activity/shopify.extension.toml.liquid
@@ -1,0 +1,27 @@
+[[extensions]]
+type = "marketing_activity"
+name = "{{ name }}"
+handle = "{{ handle }}"
+{% if uid %}uid = "{{ uid }}"{% endif %}
+title = "Your title"
+description = "Your description"
+api_path = "/replace_me"
+tactic = ""
+marketing_channel = ""
+referring_domain = ""
+is_automation = false
+
+[[extensions.preview_data.types]]
+label = "Desktop"
+value = "desktop"
+
+[[extensions.preview_data.types]]
+label = "Mobile"
+value = "mobile"
+
+[[extensions.fields]]
+name = "example_field"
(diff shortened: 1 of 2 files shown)


marketing-activity/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,27 @@
...
+[[extensions.preview_data.types]]
+label = "Desktop"
+value = "desktop"
+
+[[extensions.preview_data.types]]
+label = "Mobile"
+value = "mobile"
+
+[[extensions.fields]]
+id = "ec92dd42-8aae-4198-a6c2-20f599c82494"
```
billycai, line 23: just curious, will this error out if it's not unique?

billycai (approved):

marketing-activity/shopify.extension.toml.liquid
reemrazak: I'm guessing each extension needs to have unique fields (not across extensions) but let me tophat and double check 👀

marketing-activity/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,27 @@
+[[extensions]]
+type = "marketing_activity"
+name = "{{ name }}"
+handle = "{{ handle }}"
+{% if uid %}uid = "{{ uid }}"{% endif %}
+title = "Your title"
+description = "Your description"
+app_api_url = "https://url.com/api/execute"
+tactic = "ad"
+channel = "social"
```
billycai, line 10: I think we decided to call this `marketing_channel` instead, right?

billycai: also wondering if we should leave this and the tactic blank, otherwise I feel like people might forget to update

billycai (approved):

PR #134, opened 2024-08-30
Update validation settings templates to be consistent with the tutorial

aledalgrande:
### Background

Related https://github.com/Shopify/extensions-templates/issues/103
Corresponds to https://github.com/Shopify/shopify-dev/issues/43285

Changing to a more detailed extension in all 4 flavors to be in line with the tutorial on shopify.dev.

### Solution

Added a full solution making use of metafields and `FunctionSettings` and `Section` components.

### Checklist

- [x] I have :tophat:'d these changes
- [x] I have squashed my commits into chunks of work with meaningful commit messages

Issue #103, opened 2024-04-18
Templates improvements


aledalgrande:
Instructions here: https://docs.google.com/document/d/16-n054JAN8Ejf6lIVa4t44g3MpIVxCLRMQov9PvfOwk/edit#heading=h.mmn1j3r2z938

PR #140, opened 2024-09-25
Improvements for validation settings template

avocadomayo:
### Background

Related https://github.com/Shopify/extensions-templates/issues/103
Follows https://github.com/Shopify/extensions-templates/pull/134

### Solution

This PR does two things:

- Fix a bug with the example validation settings app
  - We previously checked `api.data.validation?.metafields` to see if a metafield definition exists. However, this field is a bit misleading such that it evaluates to `[]` when the definition does not exist OR definition exists and no value is defined. As a result, we end up creating the metafield definition on each render
  - Add a `metafieldDefinitions` query to explicitly check for existing metafield definitions, and only create a new definition if there is truly no existing definition
- Remove layout component `BlockStack`, no longer necessary since we provide the layout directly in `FunctionSettings`

### Checklist

### DIFF
--- a/validation-settings/src/ValidationSettings.liquid
+++ b/validation-settings/src/ValidationSettings.liquid
@@ -53,19 +53,18 @@ function ValidationSettings({ configuration, products }) {
 
   const onChange = async (variant, value) => {
     setErrors([]);
-    setSettings((prev) => ({
-      ...prev,
+    const newSettings = {
+      ...settings,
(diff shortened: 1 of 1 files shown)


avocadomayo:
Hey @cpeddecord 👋 I moved `applyMetafieldChange` back to `onChange` for the number field. Sorry for the mix-up 🙏

cpeddecord (approved):

PR #142, opened 2024-10-01
Add the purchase option extensions template

JulienHe:
### Background

The selling strategies is introducing 2 new admin action:
- `admin.product-purchase-option.action.render`
- `admin.product-variant-purchase-option.action.render`

We are shipping a new template so apps partners can use it and have a first starter when they want to work with these new extensions

### Solution

We are adding the new extensions to the available templates. 

### Checklist

### DIFF
--- a/admin-purchase-options-action/package.json.liquid
+++ b/admin-purchase-options-action/package.json.liquid
@@ -6,8 +6,8 @@
   "license": "UNLICENSED",
   "dependencies": {
     "react": "^18.0.0",
-    "@shopify/ui-extensions": "2024.10.x",
-    "@shopify/ui-extensions-react": "2024.10.x"
+    "@shopify/ui-extensions": "unstable",
+    "@shopify/ui-extensions-react": "unstable"
   },
   "devDependencies": {
     "@types/react": "^18.0.0",
@@ -21,7 +21,7 @@
   "version": "1.0.0",
   "license": "UNLICENSED",
   "dependencies": {
-    "@shopify/ui-extensions": "2024.10.x"
+    "@shopify/ui-extensions": "unstable"
   }
 }
 {%- endif -%}
--- a/admin-purchase-options-action/src/ProductExtension.liquid
+++ b/admin-purchase-options-action/src/ProductExtension.liquid
@@ -2,7 +2,7 @@
 import {reactExtension} from '@shopify/ui-extensions-react/admin';
 import PurchaseOptionsActionExtension from './PurchaseOptionsActionExtension';
 
-reactExtension('admin.product-purchase-option.action.render', () => (
+export default reactExtension('admin.product-purchase-option.action.render', () => (
   <PurchaseOptionsActionExtension extension="admin.product-purchase-option.action.render" />
 ));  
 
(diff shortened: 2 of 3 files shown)


JulienHe:
> We actually shouldn't merge this until the targets are in GA

Yes, we have an [epic](https://github.com/Shopify/subscriptions-app-remix/issues/1855) with what need to be shipped first.

admin-purchase-options-action/package.json.liquid
```diff
@@ -0,0 +1,27 @@
+{%- if flavor contains "react" -%}
+{
+  "name": "{{ handle }}",
+  "private": true,
+  "version": "1.0.0",
+  "license": "UNLICENSED",
+  "dependencies": {
+    "react": "^18.0.0",
+    "@shopify/ui-extensions": "unstable",
+    "@shopify/ui-extensions-react": "unstable"
```
Fionoble, line 10: These should be pointing to the latest stable calver (10-24) not to unstable

MitchLillie (changes requested):

admin-purchase-options-action/package.json.liquid
JulienHe: Thanks~ Will change.

admin-purchase-options-action/src/PurchaseOptionsActionExtension.liquid
```diff
@@ -0,0 +1,317 @@
...
+  'admin.product-variant-purchase-option.action.render';
+
+// The second argument to the render callback provides access to several useful APIs like i18n, close, and data.
+extend(PRODUCT_TARGET, (root, {i18n, close, data}) => {
+  PurchaseOptionsAction(root, {i18n, close, data});
+});
+
+extend(PRODUCT_VARIANT_TARGET, (root, {i18n, close, data}) => {
+  PurchaseOptionsAction(root, {i18n, close, data});
+});
```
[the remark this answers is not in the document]
JulienHe: Thanks for giving all these insight, I changed the templates :)

MitchLillie (approved):
LGTM!

admin-purchase-options-action/package.json.liquid
```diff
@@ -0,0 +1,27 @@
+{%- if flavor contains "react" -%}
+{
+  "name": "{{ handle }}",
+  "private": true,
+  "version": "1.0.0",
+  "license": "UNLICENSED",
+  "dependencies": {
+    "react": "^18.0.0",
+    "@shopify/ui-extensions": "2024.10.x",
```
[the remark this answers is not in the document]
MitchLillie: Correct, good catch

vividviolet (approved):

admin-purchase-options-action/src/PurchaseOptionsActionExtension.liquid
```diff
@@ -0,0 +1,297 @@
...
+
+  const updateDeliveryOption = (field, value) => {
+    setDeliveryOptions((prevOptions) => ({
+      ...prevOptions,
+      [field]: value,
+    }));
+  };
+
+  function handleSave() {
+    // This is where you can use the sellingPlanGroupsCreate and sellingPlanGroupsUpdate mutations
```
paulomarg, line 37: Will running this for products or variants change how a dev would call these mutations? If so, should we have an example of how to tell which one we're rendering?

paulomarg (approved):
Just a nit / question, but otherwise LGTM

PR #150, opened 2024-10-08
Update extend -> extension

elanalynn:
### Background

Many templates were using old `extend` method to register extensions. 

### Solution

Update to current method for registering extensions. 

`extend` -> `extension`

### Checklist

### DIFF
--- a/checkout-post-purchase/src/index.liquid
+++ b/checkout-post-purchase/src/index.liquid
@@ -133,7 +133,7 @@ import {
  * optionally allows data to be stored on the client for use in the `Render`
  * extension point.
  */
- extend("Checkout::PostPurchase::ShouldRender", async ({ storage }) => {
+extend("Checkout::PostPurchase::ShouldRender", async ({ storage }) => {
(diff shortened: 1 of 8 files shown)


checkout-post-purchase/src/index.liquid
```diff
@@ -133,7 +133,7 @@ import {
  * optionally allows data to be stored on the client for use in the `Render`
  * extension point.
  */
- extend("Checkout::PostPurchase::ShouldRender", async ({ storage }) => {
+extend("Checkout::PostPurchase::ShouldRender", async ({ storage }) => {
```
vividviolet, line 136: Let's revert this change to remove the space

elanalynn: @vividviolet  I saw this as well, but I think the space was off before. It wasn't aligned with the functions around it.

vividviolet: Oops, you're right. Ignore me

PR #158, opened 2024-10-22
Create flow-trigger-discovery-callback template

mikee649:
### Background

Partial: https://github.com/Shopify/flow/issues/27653

Adds new template for `trigger_discovery_callback`

### Checklist


flow-trigger-discovery-callback/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,5 @@
+[[extensions]]
+name = "{{ name }}"
+handle = "{{ handle }}"
+type = "trigger_discovery_callback"
```
steve-low, line 4: This makes the extension name a publicly visible field in that developers will see it and refer to it by this value. We'd have to update the documentation and public naming from `webhook` to `callback`.

This puts some more weight on the rename. Let's discuss this a little more.

mikee649: Wasn't a conscious decision, mostly just because the original `trigger_discovery_webhook` didn't have it. But it looks like it's there for both `flow_trigger` and `flow_action`, so it's probably worth putting back.

mikee649: We've agreed offline to go with `flow_trigger_discovery_callback`

flow-trigger-discovery-callback/shopify.extension.toml.liquid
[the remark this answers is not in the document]
mikee649: 👍

flow-trigger-discovery-callback/shopify.extension.toml.liquid
mikee649: re-agreed to go with `flow_trigger_lifecycle_callback`

PR #160, opened 2024-10-25
Update directory name, details and add conditional checks for flavors

elanalynn:
### Background

I misnamed the folder in the initial PR. 

### Solution

This PR update `condtional-admin-extension` to `condtional-admin-extension`, clarifies some language, updates the registration method name and creates a conditional check in the `package.json`

### Checklist


conditional-action-extension/README.md.liquid
```diff
@@ -6,6 +6,6 @@ Admin action extensions enable app developers to build custom functionality into
...
 
-## Admin data extensions to conditionally control action extension triggers
+## Admin data extension to conditionally control action extension triggers
 
-Admin data extensions run in the browser like UI extensions but do not return any UI. Instead, an admin data extension returns specific data, as permitted by the contract with the extension point that it runs in. Admin action condition extensions are a specific type of admin data extension used to control the visibility of an admin action trigger in the "More actions" menu on a given resource page. They are expected to return a data object that determines the visibility of a related admin action extension. 
+Admin data extensions run in the browser like UI extensions but do not return any UI. Instead, an admin data extension returns specific data, as permitted by the contract with the extension point that it runs in. Admin action "should render" extensions are a specific type of admin data extension used to control the visibility of an admin action trigger in the "More actions" menu on a given resource page. They are expected to return a data object that determines the visibility of a related admin action extension.
```
belalsj, line 11: `Admin action "should render" extensions` - I wonder if this is the name we should use. Seems wordy. I get that we're moving away from `.condition` but they're still sort of "condition controllers". Thoughts?

elanalynn: Hmm... I don't know. @AaronSpringut do you have any thoughts about the language here?

AaronSpringut: Let's call them "conditional rendering extensions", which are specified with the "should render" property.

PR #163, opened 2024-10-30
Make rust first in function templates

davejcameron:
### Background

Rust is our default language choice

### Solution

### Checklist


craigmichaelmartin (approved):
Looks great!

PR #165, opened 2024-11-03
Push templates.json to CDN when it changes

amcaplan:
### Background

We would like:

1. To continue to have this repo be the source of truth for templates. 1 new/updated template, 1 repo to update.
2. To use our own CDN to propagate the file, as GitHub has a [rate limit](https://github.com/github/docs/issues/8031#issuecomment-881427112) that might be hit under certain circumstances.

### Solution

We continue committing here, and set up a job that automatically copies the file to an internal repo which will trigger upload to CDN.

We thought of a few other options:

1. Move the source of truth to an internal repo which will be uploaded to CDN - in which case we lose the 1-PR atomicity
2. Use a public CDN like jsdeliver [which can reflect GitHub files](https://www.jsdelivr.com/?docs=gh)

However, it seemed like using our own CDN is the right move, reducing external points of failure. And we prefer a solution that keeps the 1-PR atomicity.

### Checklist

I deleted the checklist since this is a GitHub action PR. So we have to test post-merge.

### DIFF
--- a/.github/actions/push-templates-json.yml
+++ b/.github/actions/push-templates-json.yml
@@ -1,0 +1,29 @@
+---
+name: Push Templates JSON
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'templates.json'  # only trigger when this file changes
+  workflow_dispatch: # allow manual runs
+
+jobs:
+  push-templates:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Push templates
+        run: npm run push-templates-json
--- a/bin/push-templates-json.js
+++ b/bin/push-templates-json.js
@@ -1,0 +1,57 @@
+#!/usr/bin/env node
+process.removeAllListeners('warning');
+
+// ES Modules
+import {createRequire} from 'module'
+import {fileURLToPath} from "node:url"
+import * as path from "pathe"
+
+// CJS Modules
+const require = createRequire(import.meta.url)
+const {readFile} = require('fs-extra')
+const {Octokit} = require('@octokit/core')
+const {createPullRequest} = require("octokit-plugin-create-pull-request")
+const colors = require('ansi-colors')
+
+const fileToPush = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "../templates.json")
+console.log("Opening a PR to update the templates.json file on CDN")
+
+const files = {
+  "static-24h/cli/extensions/templates.json": await readFile(fileToPush)
+}
+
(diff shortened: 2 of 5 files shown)


bin/push-templates-json.js
```diff
@@ -0,0 +1,57 @@
...
+const octokit = new OctokitWithPlugin({
+  auth: process.env.GITHUB_TOKEN,
+});
+
+const response = await octokit
+  .createPullRequest({
+    owner: "Shopify",
+    repo: "static-cdn-assets",
+    title: `Shopify CLI - update extension templates - ${new Date().toISOString().split('T')[0]}`,
+    body: `We are updating the extension templates available for generating in apps via Shopify CLI`,
```
MitchDickinson, line 33: Should we include some context here that this is an automated PR? Will it be obvious?

bin/push-templates-json.js
```diff
@@ -0,0 +1,57 @@
...
+    forceFork: false,
+    changes: [
+      {
+        files,
+        commit: `chore: update extension templates`,
+      },
+    ],
+  })
+
+// eventually activate this once we have confidence in the PR
```
MitchDickinson, line 46: I'm not sure automatically merging will ever be the right thing here, so maybe we should remove this and try with the manual process. If we find it's working really well and we want to automate it, we can always add this back later.

bin/push-templates-json.js
MitchDickinson, line 1: Not really much logic but i wonder if this file should have a test associated with it?

bin/push-templates-json.js
amcaplan: This is exactly what we do for Homebrew, but I can add in something a bit more explicit.

bin/push-templates-json.js
amcaplan: This is really simple code, the main issue is whether the pieces work together. I'd be more inclined just to be sure some alarm will be set off in case this fails.

bin/push-templates-json.js
amcaplan: I think that making this automatic is the key to making this really the source of truth. Everything will be propagated automatically once a change is made.

I'd like to understand better the case for not auto-merging.

.github/actions/push-templates-json.yml
```diff
@@ -0,0 +1,29 @@
+---
+name: Push Templates JSON
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'templates.json'  # only trigger when this file changes
```
ajshepley, line 9: Is this file going to be at the root of the directory? I think the path specified by `paths` is a path glob relative to root: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths

amcaplan: Yes, it's right here: https://github.com/Shopify/extensions-templates/blob/main/templates.json

amcaplan:
Closing in favor of https://github.com/Shopify/static-cdn-assets/pull/1114

PR #182, opened 2024-12-05
Update Admin Link template to align with Action Links

cameronbarker:
### Background

Admin Link's previously have used `title` as the text of the link. Going forward we've decided to align with action extensions which uses the extension `name` property as the text of the link.

### Checklist

### DIFF
--- a/admin-link/locales/fr.json.liquid
+++ b/admin-link/locales/fr.json.liquid
@@ -1,3 +1,3 @@
 {
-  "text": "link-extension"
+  "name": "Texte du lien d'extension"
 }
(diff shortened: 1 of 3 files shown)


admin-link/locales/fr.json.liquid
```diff
@@ -1,3 +1,3 @@
 {
-  "text": "link-extension"
+  "name": "link-extension"
```
vividviolet, line 2: Can we update these to a more realistic one? This will look super similar to the handle `admin-link` and isn't very helpful. How about `"Product details"`?

vividviolet (approved):

AaronSpringut (approved):

PR #194, opened 2025-01-08
Update ui-extensions versions in templates

elanalynn:
### Background

We just released the [`2025.1.0`](https://www.npmjs.com/package/@shopify/ui-extensions) versions of `ui-extensions` and `ui-extensions-react`. This updates our templates to use the latest versions. 

### Solution

Update versions to latest

### Checklist

### DIFF
--- a/admin-action/package.json.liquid
+++ b/admin-action/package.json.liquid
@@ -6,8 +6,8 @@
   "license": "UNLICENSED",
   "dependencies": {
     "react": "^18.0.0",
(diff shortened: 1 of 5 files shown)


MitchLillie (approved):
Approved but we should also update print actions

elanalynn:
> Approved but we should also update print actions

@MitchLillie  I left those out intentionally, since we will be adding conditional rendering to print actions in the next few days, but I suppose that will be a different template.

PR #200, opened 2025-01-25
Skeleton channel config template

dsrw:
### Background

As part of the [Common channel listing management](https://vault.shopify.io/gsd/projects/43512-Common-channel-listing-management-experience) project we're building a [Channel config extension](https://github.com/Shopify/shopify/pull/573560) to allow configuring product requirements for sales channels. This is a skeleton version of the extension template, to be filled out in future PRs.

This will be gated behind the [common_channel_listing_management_experience](https://github.com/Shopify/business-platform/pull/24187/files) org feature flag.

### Checklist

### DIFF
--- a/channel-config/README.md.liquid
+++ b/channel-config/README.md.liquid
@@ -1,0 +1,4 @@
+## Extension Configuration
+
+The Channel Config Extension allows channel apps to provide channel details and
+product requirements needed by Shopify's Common Listing Management Experience.
--- a/channel-config/shopify.extension.toml.liquid
+++ b/channel-config/shopify.extension.toml.liquid
@@ -1,0 +1,8 @@
+[[extensions]]
+type = "channel_config"
+handle = "{{ handle }}"
+name = "{{ name }}"
+{% if uid %}uid = "{{ uid }}"{% endif %}
+
+[channel_requirements]
+channel_handle = "Shop"
--- a/templates.json
+++ b/templates.json
@@ -1156,5 +1156,22 @@
         "path": "admin-link"
       }
     ]
+  },
+  {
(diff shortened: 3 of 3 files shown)


channel-config/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,7 @@
+[[extensions]]
+type = "channel_config"
```
pt2pham, line 2: I believe `name = ` is also a required property if I'm not mistaken, so you can add that as well

channel-config/shopify.extension.toml.liquid
```diff
@@ -0,0 +1,7 @@
...
+handle = "{{ handle }}"
+{% if uid %}uid = "{{ uid }}"{% endif %}
+
+[[extensions.channel_requirements]]
+handle = "Shop"
```
pt2pham, line 7: `handle` is kind of a loaded term especially with extension handles and app handles - should this just be `channel` to be really clear?

pt2pham (approved):
Very minor comments

channel-config/README.md.liquid
```diff
@@ -0,0 +1 @@
+## Extension Configuration
```
Catherine-Fu, line 1: nit: can we add a brief introduction for this channel config extension?

Catherine-Fu (approved):

channel-config/shopify.extension.toml.liquid
dsrw: I renamed it to `channel_handle`. This will ultimately end up in the existing `Channel#handle` field, so I'd like to keep `handle` in there if possible.

PR #201, opened 2025-01-31
Remove post purchase detail extensions from POS

vctrchu:
### Background

Removes the post-purchase templates as they are not crucial components in getting your first extension running.


NathanJolly (approved):
Thank you!

PR #206, opened 2025-02-27
feat(products): Update product configuration extensions to include GraphQL Query sample and support React 18 and latest UI extensions

kmdavis:
### Background

This PR:
- adds graphql querying for product/variant + components (in place of the existing (& deprecated) `data.product`/`data.variant` json blob that is passed in.
- includes a bare minimal list of components
- updates dependencies

### Checklist

### DIFF
--- a/product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
+++ b/product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
@@ -2,38 +2,186 @@
 import {
   reactExtension,
   useApi,
+  BlockStack,
   Text,
 } from '@shopify/ui-extensions-react/admin';
+import {useState, useEffect} from 'react';
 
 // The target used here must match the target used in the extension's toml file (./shopify.extension.toml)
-{% if flavor contains "typescript" %}
-export default reactExtension<any>('admin.product-details.configuration.render', () => <App />);
-{% else %}
 export default reactExtension('admin.product-details.configuration.render', () => <App />);
-{% endif %}
+
 function App() {
   {% if flavor contains "typescript" %}
   const {extension: {target}, i18n} = useApi<'admin.product-details.configuration.render'>();
   {% else %}
   const {extension: {target}, i18n} = useApi();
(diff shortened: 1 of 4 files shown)


product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
```diff
@@ -2,38 +2,186 @@
...
+          bundleComponents: bundleComponents.nodes.map(({componentProduct}: any) => ({
+          {% else %}
+          bundleComponents: bundleComponents.nodes.map(({componentProduct}) => ({
+          {% endif %}
+            ...componentProduct
+          }))
+        })
+      }
+    })
+  }, [productId]);
```
elanalynn, line 90: Should `query` be included in the dependencies?

product-configuration-extension/src/ProductVariantDetailsConfigurationExtension.liquid
```diff
@@ -2,38 +2,186 @@
...
+          productVariantComponents: productVariantComponents.nodes.map(({productVariant}: any) => ({
+          {% else %}
+          productVariantComponents: productVariantComponents.nodes.map(({productVariant}) => ({
+          {% endif %}
+            ...productVariant
+          }))
+        })
+      }
+    })
+  }, [productVariantId])
```
elanalynn, line 90: Should `query` be included here too?

elanalynn (approved):
🎩 I generated a TS template. Looks good, with just a couple comments about the deps arrays.

product-configuration-extension/src/ProductDetailsConfigurationExtension.liquid
kmdavis: oops, yeah, it probably should be

kmdavis:
> For liquid files is it possible to get a preview/screenshot? Asking because I see js/ts code

Screenshot 2025-02-27 at 11 05 31 AM

Screenshot 2025-02-27 at 11 07 05 AM

PR #208, opened 2025-03-13
Add discount function

devisscher:
### Background

Adds discount function template with Rust, JavaScript, TypeScript and Wasm support.

### Solution

### Checklist


jonathanhamel4 (approved):
lgtm. Do we need to keep the wasm target there or remove it?

PR #217, opened 2025-04-09
Update checkout UI templates for 2025-07 RC

kumar303:
Fixes https://github.com/Shopify/checkout-web/issues/44281

This updates the template for newly generated Checkout UI extensions.

# 🎩 Instructions

Make sure `pnpm` is up to date

Use [this CLI snapshot](https://github.com/Shopify/cli/pull/5633)

```
pnpm i -g @shopify/cli@0.0.0-snapshot-20250410123718 --@shopify:registry=https://registry.npmjs.org
```

Check the version

```
% shopify --version
@shopify/cli/0.0.0-snapshot-20250410123718 darwin-arm64 node-v23.10.0
```

Create an app. Select _Remix_ and _TypeScript_.

```
shopify app init
```

Change into the app directory and generate an extension with this template

```
POLARIS_UNIFIED=true shopify app generate extension --clone-url https://github.com/Shopify/extensions-templates#km-checkout-templates   
```

Run the app

```
shopify app dev
```

Here's what it should look like

<details>
<summary>📸 Screenshot</summary>

Screenshot 2025-04-10 at 5 59 59 PM

</details>

### Background

### Solution

### Checklist

### DIFF
--- a/checkout-extension/package.json.liquid
+++ b/checkout-extension/package.json.liquid
@@ -1,27 +1,26 @@
-{%- if flavor contains "react" -%}
 {
   "name": "{{ handle }}",
   "private": true,
   "version": "1.0.0",
   "license": "UNLICENSED",
+{%- if flavor contains "preact" %}
+  "dependencies": {
+    "preact": "^10.10.x",
+    "@shopify/ui-extensions": "2025-07-rc"
+  }
+{%- elsif flavor contains "react" %}
   "dependencies": {
     "react": "^18.0.0",
-    "@shopify/ui-extensions": "2025.1.x",
-    "@shopify/ui-extensions-react": "2025.1.x",
+    "@shopify/ui-extensions": "2025.4.x",
+    "@shopify/ui-extensions-react": "2025.4.x",
     "react-reconciler": "0.29.0"
   },
   "devDependencies": {
     "@types/react": "^18.0.0"
   }
-}
-{%- else -%}
-{
-  "name": "{{ handle }}",
-  "private": true,
-  "version": "1.0.0",
-  "license": "UNLICENSED",
+{%- else %}
   "dependencies": {
-    "@shopify/ui-extensions": "2025.1.x"
+    "@shopify/ui-extensions": "2025.4.x"
   }
+{%- endif %}
 }
-{%- endif -%}
--- a/checkout-extension/shopify.extension.toml.liquid
+++ b/checkout-extension/shopify.extension.toml.liquid
@@ -3,7 +3,11 @@
 
 # The version of APIs your extension will receive. Learn more:
 # https://shopify.dev/docs/api/usage/versioning
-api_version = "2025-01"
+{%- if flavor contains "preact" %}
+api_version = "2025-07"
+{%- else %}
+api_version = "2025-04"
+{%- endif %}
 
 [[extensions]]
 name = "{{ name }}"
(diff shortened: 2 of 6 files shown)


checkout-extension/package.json.liquid
vividviolet, line 1: For consistency can you use the template [here](https://github.com/Shopify/extensions-templates/blob/63a1d061e8799633257d21c845a59b4068648b1c/admin-action/package.json.liquid#L1C1-L24C4)? I verified that it produces the right JSON with proper line breaks

checkout-extension/shopify.extension.toml.liquid
```diff
@@ -3,7 +3,11 @@
 
 # The version of APIs your extension will receive. Learn more:
 # https://shopify.dev/docs/api/usage/versioning
-api_version = "2025-01"
+{% if flavor contains "preact" -%}
+api_version = "2025-07"
+{%- else -%}
```
vividviolet, line 8: Might be one too many `-`?
```suggestion
{% else %}
```

I verified that this one works as expected: https://github.com/Shopify/extensions-templates/blob/63a1d061e8799633257d21c845a59b4068648b1c/admin-action/shopify.extension.toml.liquid#L1

checkout-extension/shopify.extension.toml.liquid
```diff
@@ -3,7 +3,11 @@
...
+api_version = "2025-04"
+{%- endif %}
```
vividviolet, line 10:
```suggestion
{% endif %}
```

checkout-extension/tsconfig.json.liquid
```diff
@@ -0,0 +1,27 @@
+{%- if flavor contains "preact" -%}
+{
+  // This tsconfig.json file is only needed to inform the IDE.
```
vividviolet, line 3: We actually should remove this comment. The tsconfig is now required to build preact extensions properly and also to do the type checks.

Can you use the template [here](https://github.com/Shopify/extensions-templates/blob/63a1d061e8799633257d21c845a59b4068648b1c/admin-action/tsconfig.json.liquid#L1)?

checkout-extension/src/Checkout.liquid
```diff
@@ -1,4 +1,66 @@
...
+  // 3. Render a UI
+  return (
+    <s-banner title="{{ name }}">
+      <s-stack gap="base">
+        <s-text>
+          {shopify.i18n.translate("welcome", {
+            target: <s-text type="emphasis">{shopify.extension.target}</s-text>,
+          })}
+        </s-text>
+        <s-button onClick={handleClick}>
```
kumar303, line 33: `<s-checkbox>` doesn't work in Checkout yet 😓

jamesvidler (approved):
Tophatted and everything works as expected. We can definitely improve the actual extension code (like using a checkbox) and using real hooks (once available). But I think we can follow up with that after.

How are we going to be documenting to the public how to generate extensions using `preact`? Are we going to keep the `POLARIS_UNIFIED=true` flag to handle this so partners can opt into the remote-dom + release candidate? Otherwise, everyone else will continue to get `2025-04` with options between React js/ts, and vanilla js/ts?

I'm approving because i'm assuming you can apply some of the recommended template updates to solve the code formatting.

vividviolet:
> How are we going to be documenting to the public how to generate extensions using `preact`? Are we going to keep the `POLARIS_UNIFIED=true` flag to handle this so partners can opt into the remote-dom + release candidate? Otherwise, everyone else will continue to get `2025-04` with options between React js/ts, and vanilla js/ts?

Yes. The flag is only kept in all CLI versions leading up to 2025-07 becoming stable. After 2025-07 is stable subsequent CLI releases should use remote-dom templates by default.

PR #219, opened 2025-04-11
Fix incorrect property in <s-banner> of Checkout template

kumar303:
### Background

This is a follow-up to https://github.com/Shopify/extensions-templates/pull/217

The `title` heading was invalid for `<s-banner>`

### Solution
🔧 

### Checklist


simontaisne (approved):
💯

PR #260, opened 2025-05-23
Update title / heading props for admin action and banner

kyledurand:
Update incorrect props


vividviolet (approved):
thanks for fixing this!

PR #264, opened 2025-05-29
Fix path for delivery option discounts

davejcameron:
### Background

The URL was incorrect to the input query which was preventing the function from being uploaded when first created.

### Solution

Point to the input query

### Checklist


lizkenyon (approved):
👌

PR #270, opened 2025-06-05
Added encryption settings and location mappings url

DavidHou9:
### Background

### Solution

### Checklist


IP1102 (approved):

terminal-management/shopify.extension.toml.liquid
```diff
@@ -3,6 +3,8 @@ name = "{{ name }}"
 type = "terminal_management"
 
 terminals_url = "https://api.paymentprovider.com/terminals"
-terminal_status_url = "https://api.paymentprovider.com/terminal/status"
+terminal_details_url = "https://api.paymentprovider.com/terminal/details"
 # This is the page on your website where merchants can go to diagnose problems.
-terminal_diagnostic_page_url = "https://paymentprovider.com/help"
+terminal_diagnostic_page_url = "https://api.paymentprovider.com/terminal/setup_help"
+terminal_encryption_settings_url = "https://api.paymentprovider.com/terminal/encryption_settings"
+provider_location_mappings_url = "https://api.example.com/location_mappings"
```
kerfootj, line 10: Is the `provider` prefix necessary? All the urls point to the provider so `location_mappings_url` is probably sufficient.

kerfootj (approved):
Just a comment on the `provider_location_mappings_url` name. Either way I think this is good.

terminal-management/shopify.extension.toml.liquid
```diff
@@ -3,6 +3,8 @@ name = "{{ name }}"
 type = "terminal_management"
 
 terminals_url = "https://api.paymentprovider.com/terminals"
-terminal_status_url = "https://api.paymentprovider.com/terminal/status"
+terminal_details_url = "https://api.paymentprovider.com/terminal/details"
 # This is the page on your website where merchants can go to diagnose problems.
-terminal_diagnostic_page_url = "https://paymentprovider.com/help"
+terminal_diagnostic_page_url = "https://api.paymentprovider.com/terminal/setup_help"
+terminal_encryption_settings_url = "https://api.paymentprovider.com/terminal/encryption_settings"
```
cpilegard, line 9: I can't find an example of what we expect of this endpoint, do you have a link to that? I'm just looking to make sure "encryption settings" makes sense for the naming.

cpilegard: Thanks. I think this is fine for Adyen but there's probably more to discuss there before this goes public.

cpilegard (approved):

PR #276, opened 2025-06-20
Add Test cert in the credit card extension config by default

manefz:
Fix the template so the CLI does not automatically fail on `shopify app dev/deploy` due to missing cert. config

### Background

N/A

### Solution

Decided to include the Test Certificate by default in the template. This is an acceptable fallback initially when registering a new credit card extension.

### Checklist

### DIFF
--- a/payments-app-extension-credit-card/shopify.extension.toml.liquid
+++ b/payments-app-extension-credit-card/shopify.extension.toml.liquid
@@ -24,7 +24,7 @@ supports_3ds = false
 supports_installments = false
 supports_deferred_payments = false
 test_mode_available = true
-encryption_certificate_fingerprint = ""
+encryption_certificate_fingerprint = "Test Certificate"
 
 [[extensions.targeting]]
 target = "payments.credit-card.render"


payments-app-extension-credit-card/shopify.extension.toml.liquid
```diff
@@ -24,7 +24,7 @@ supports_3ds = false
 supports_installments = false
 supports_deferred_payments = false
 test_mode_available = true
-encryption_certificate_fingerprint = ""
+encryption_certificate_fingerprint = "Test Certificate"
```
delongGao, line 27: Question: does the content of this fingerprint need to be in any specific format for the config to work?

delongGao (commented):
Change looks good to me, thanks! 👍  I'm not very familiar with this repo, just to confirm: do we have anyway to test this default config with the template?

payments-app-extension-credit-card/shopify.extension.toml.liquid
manefz: Yeah for the CLI validation to pass it needs to be a string of 1 character at least, ie. not empty . That is only for `shopify app generate extension`
However we have stricter validation on the plugins during `deploys and releases` `shopify app deploy`.
This validation is handled by HandleEncryptionCertificatePlugin

PR #282, opened 2025-07-03
Fix order routing location rule extension template

vividviolet:
### Background

This PR put back support for different React, and vanilla JS in the OrderRoutingLocationRule template.

### Testing

1. Run `shopify app generate extension --clone-url https://github.com/Shopify/extensions-templates#Fix_Order_Routing_Location_Rule_extension_template` 
1. Verify you can generate an "Order routing location rule - UI Extension" 

### Checklist


jbalsas (approved):
Couldn't 🎩 because CLI kept refusing to log me in with the right organization, but changes make sense to me, I think!

PR #294, opened 2025-07-08
Test CI rust clippy

tjoyal:
### Background

### Solution

### Checklist


rust-toolchain.toml
adampetro, line 1: With this change, we should be able to remove [this step](https://github.com/Shopify/extensions-templates/blob/3acbf9cfcd212050d553560a9945e914c7edb442/.github/workflows/validate-rust-functions.yml#L18-L19) as well as [this](https://github.com/Shopify/extensions-templates/blob/3acbf9cfcd212050d553560a9945e914c7edb442/.github/workflows/validate-rust-functions.yml#L30-L31) from the Rust workflow as the components and target will be installed automatically

PR #307, opened 2025-08-13
[Discounts] bugfix: remove the extraneous applyExtensionMetafieldChange

iphipps:
### Background

applyExtensionMetafield is called twice, on `FunctionSettings onSave` and `Form onSubmit`.  The result is that the input value is not updated when we save the discount with the ui extension.

### Solution

I tested using only Form onSubmit, but that didn't work.  FunctionSettings onSave is required.

### Checklist


markusvoigt:
This fix appears to work for me: https://screenshot.click/15-34-0q74k-n15fr.mp4

PR #309, opened 2025-08-22
Adjust discount function settings metafield dependency array

iphipps:
### Background

Currently when you save a discount using this extension, the old metadata appears in the form fields after the save.  The data.metafields in the dependency array is an array which is compared by reference, not by value.

### Solution

This uses JSON.stringify to convert the array to a string so react can compare by value more easily.  There is some performance risk with this depending on how much data is in the array, but for the purposes of this demo, for which we use 4 fields on one object in the array, I think it's ok.  Developers should be able to figure out a more reliable way to determine that they need to memoize again, which should be straight forward depending on the data.
### Checklist

### DIFF
--- a/discount-details-function-settings-block/src/DiscountFunctionSettings.liquid
+++ b/discount-details-function-settings-block/src/DiscountFunctionSettings.liquid
@@ -201,6 +201,7 @@
   function useExtensionData() {
     const { applyMetafieldChange, i18n, data, resourcePicker, query } =
       useApi(TARGET);
+    const metafields = data?.metafields ? JSON.stringify(data.metafields) : '';
     const metafieldConfig = useMemo(
       () =>
         parseMetafield(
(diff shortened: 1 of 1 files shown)


discount-details-function-settings-block/src/DiscountFunctionSettings.liquid
```diff
@@ -201,14 +201,15 @@
   function useExtensionData() {
     const { applyMetafieldChange, i18n, data, resourcePicker, query } =
       useApi(TARGET);
+    const metafields = data?.metafields ? JSON.stringify(data.metafields) : '';
```
loic-d, line 204: It feels weirds to force the consumer to do that. This makes me wondering if there was a recent change that made `metafields` being mutated in place (same object reference, different contents), which is not forcing an update of the dependency (React checks dependencies with `Object.is`).

I'm not super familiar with how the data is exposed through the API (`launchOptions.context?.api`) but I'm wondering if we should look there?

For example, this could be relevant: https://github.com/shop/world/pull/121608/files#r2240572620

PR #315, opened 2025-09-16
Update to use latest `shopify_function` crate

jeffcharles:
### Background

The new version of the Function Wasm API does not support using WASI and the upcoming `shopify_function` Rust crate has a new logging API.

### Solution

- Update the build targets to use `wasm32-unknown-unknown` instead of `wasm32-wasip1`
- Update `process::exit(1)` to `process::abort()`
- Update `eprintln!` to `log!`

### Checklist

### DIFF
--- a/.github/workflows/validate-rust-functions.yml
+++ b/.github/workflows/validate-rust-functions.yml
@@ -28,5 +28,5 @@ jobs:
         run: cargo clippy -- -D warnings
       - name: Run tests
         run: cargo test
-      - name: Build with wasm32-wasip1 target
-        run: cargo build --release --target wasm32-wasip1
+      - name: Build with wasm32-unknown-unknown target
+        run: cargo build --release --target wasm32-unknown-unknown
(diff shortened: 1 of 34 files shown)


lopert (approved):
I 🎩 'ed this by generating a cart checkout validation function.

```
shopify app generate extension -u https://github.com/Shopify/extensions-templates.git\#jc.update-to-in-memory-io-api
```

When trying to deploy it ran into the following "issues":

---

Rustup target was missing
```
|  error[E0463]: can't find crate for `core`                                                                                  │
│    |                                                                                                                        │
│    = note: the `wasm32-unknown-unknown` target may not be installed                                                         │
│    = help: consider downloading the target with `rustup target add wasm32-unknown-unknown`                                  │
```
Solved with
```
rustup target add wasm32-unknown-unknown
```
---

Then I ran into
```
unknown import: `shopify_function_v2::shopify_function_input_get` has not been defined
```
Solved by updating to Shopify CLI 3.85
```
pnpm install -g @shopify/cli@latest
````

---

I also ran the resulting WASM through the latest version of function-runner.

PR #318, opened 2025-09-18
Remove channel specifications from channel-config extension

maurizio-cucchiara:
### Background

Removed the `specifications` field from the Shopify extension configuration template as it can easily inferred by the specifications folder content

### Checklist


Fernanda-Montiel:
Can we also update the field `automaticProductFeedManagement`? The validation is asking for productFeedManagement with no-boolean values such as "automatic"

Fernanda-Montiel (approved):

PR #327, opened 2025-09-29
Fix type issues by using event.currentTarget instead of event.target

vividviolet:
### Background

Fix type issues
- Updated event handling to use `event.currentTarget.value` instead of `event.target.value` for accessing form field values.
- Changed the initial state values for frequency and discount from numbers to strings in PurchaseOptionsActionExtension to reflect the expected value for s-number-field
- Updated the function settings component to properly use `onSubmit` event with `waitUntil` instead of `onSave`

### Checklist


vividviolet:
### Merge activity

* **Sep 29, 3:26 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/extensions-templates/327).

PR #331, opened 2025-09-30
Use an async arrow function for checkout

kumar303:
### Background

Follow the pattern of Admin and match documentation examples in Checkout.

### Solution

Use an async arrow function

### Checklist


lrsterrett (approved):
👍

PR #332, opened 2025-10-02
Preact discount settings bug fixes

david-hollenbeck-shopify:
### Background

Howdy!

I was putting together a demo this morning and ran into a few bugs after generating a new extension-only app with a discount function + Discount function settings ui (CLI `3.85.4`)

- infinite `useEffect` render loop
- select menu not working

Both quick fixes that I've already tested for the demo, dropping them here in case it hasn't been flagged yet.

### Solution

- infinite render is resolve by apply `useMemo` (same dependency as react template)
- updated select menu to use <s-option> from polaris web components

### Checklist


vividviolet (approved):
Makes sense to me!

PR #387, opened 2026-04-02
Add app.intent.render template

JoviDeCroock:
### Background

This adds a new targetted example for `admin.app.intent.render`
### Checklist

### DIFF
--- a/admin-intent-render/src/IntentExtension.liquid
+++ b/admin-intent-render/src/IntentExtension.liquid
@@ -79,9 +79,9 @@ function Extension() {
               value={layout}
               onChange={(event) => setLayout(event.currentTarget.value)}
             >
-              <option value="single-column">Single column</option>
-              <option value="two-column">Two column</option>
-              <option value="hero-image">Hero image</option>
+              <s-option value="single-column">Single column</s-option>
+              <s-option value="two-column">Two column</s-option>
+              <s-option value="hero-image">Hero image</s-option>
             </s-select>
           </s-stack>
           <s-stack>
--- a/admin-intent-render/src/IntentExtension.liquid
+++ b/admin-intent-render/src/IntentExtension.liquid
@@ -1,10 +1,15 @@
 import "@shopify/ui-extensions/preact";
 import {render} from 'preact';
-import {useState} from 'preact/hooks';
+import {useEffect, useState} from 'preact/hooks';
 
 export default async () => {
-  shopify.tools.register('design_email', designEmail);
   render(<Extension />, document.body);
+};
+
+const defaultValues = {
+  backgroundColor: '#ffffff',
+  layout: 'single-column',
+  fontFamily: '',
 };
 
 function Extension() {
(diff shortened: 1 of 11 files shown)


kbav (commented):
WIP review -- In process of tophatting

Also have a couple comments/questions

kbav (commented):
## Tophat success

Was able to get the CLI to generate the extension. It required:
- having the branch locally and using right CLI env variable
- modifying this PR's code ([now reverted back to your original commit](https://github.com/Shopify/extensions-templates/pull/387#discussion_r3034068473)) to point the the snapshot version
- `@shopify/ui-extensions` snapshot to be ready
- using `--clone-url` flag

...but it worked!

Happy to approve once everything's ready to go

Image

admin-intent-render/src/IntentExtension.liquid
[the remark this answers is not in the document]
kbav: This question is still valid, but not blocking

kbav (approved):
1 non-blocking comment, LGTM and thank you for the iteration!

admin-intent-render/src/IntentExtension.liquid
```diff
@@ -0,0 +1,111 @@
...
+  );
+  const [layout, setLayout] = useState(
+    typeof data?.layout === 'string' ? data.layout : 'single-column',
+  );
+  const [fontFamily, setFontFamily] = useState(
+    typeof data?.font_family === 'string' ? data.font_family : '',
+  );
+
+  async function resolveIntent(callback) {
+    if (!intents.response) {
```
vividviolet: I don't think we need this conditional because this target will always have it. Can we remove?

admin-intent-render/src/IntentExtension.liquid
```diff
@@ -0,0 +1,111 @@
...
+      setApplying(false);
+      setError('Could not resolve the app intent.');
+    }
+  }
+
+  const apply = () => {
+    setApplying(true);
+    resolveIntent(async (response) => {
+      const result = await designEmail({
+        background_color: backgroundColor,
```
vividviolet: nit: I don't think there's anything forcing us to use camel case so we can just make the schema use pascal casing to simplify it

admin-intent-render/src/IntentExtension.liquid
```diff
@@ -0,0 +1,111 @@
...
+        layout,
+        font_family: fontFamily,
+      });
+      return response.ok(result);
+    });
+  };
+
+  return (
+    <s-form onSubmit={apply}>
```
vividviolet: I think we're missing a `onReset` that's supposed to reset the form values back to the last saved values

admin-intent-render/src/IntentExtension.liquid
```diff
@@ -0,0 +1,111 @@
...
+            {applying ? 'Applying...' : 'Apply design'}
+          </s-button>
+        </s-stack>
+      </s-stack>
+    </s-form>
+  );
+}
+
+async function designEmail(input) {
+  console.log('Designing email', input);
```
vividviolet: Can we actually hook this up to update the form values? It would be a cool end-to-end demo

templates.json
```diff
@@ -2122,5 +2122,26 @@
       "99125067"
     ],
     "minimumCliVersion": "3.90.0"
+  },
+  {
+    "identifier": "admin_intent_render",
```
vividviolet: I wonder if we should just name this `admin_intent`? Most of our other ui extensions are this way. Admin links gets the `_link` suffix so i think it should be clear enough

templates.json
```diff
@@ -2122,5 +2122,26 @@
...
+      {
+        "name": "Preact",
+        "value": "preact",
+        "path": "admin-intent-render"
+      }
+    ],
+    "organizationExpFlags": [
+      "99125067"
+    ],
+    "minimumCliVersion": "3.90.0"
```
vividviolet: This should be the upcoming release `3.94.0`. You can change the `CLI_KIT_VERSION` on the CLI for testing

admin-intent-render/src/IntentExtension.liquid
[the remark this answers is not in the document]
JoviDeCroock: No other source files use jsx, they use liquid

admin-intent-render/src/IntentExtension.liquid
```diff
@@ -0,0 +1,124 @@
...
+
+  const [applying, setApplying] = useState(false);
+  const [error, setError] = useState('');
+
+  const [backgroundColor, setBackgroundColor] = useState(initialValues.backgroundColor);
+  const [layout, setLayout] = useState(initialValues.layout);
+  const [fontFamily, setFontFamily] = useState(initialValues.fontFamily);
+
+  useEffect(() => {
+    shopify.tools.register('design_email', async (input) => {
```
vividviolet: Can you return the `unregister` callback from `shopify.tools.register` in the `useEffect` just for best practice

vividviolet (approved):
Just a tiny thing to fix and then it's good to go

templates.json
adambarrus: Agree that we should not have "render" in the naming. 

As per discussion in Slack, we should have this in the CLI:

```
Sidekick
   App data
   App action
   App action link
```

PR #393, opened 2026-04-15
Add order-attribution-config extension template

Catherine-Fu:
## Summary
- Add `order-attribution-config/` template directory for the `order_attribution_config` extension type
- Add entry to `templates.json` under "Sales channels" group
- Enables `shopify app generate extension` to scaffold order attribution config extensions

## Template structure
```
order-attribution-config/
├── shopify.extension.toml.liquid
├── icons/
│   └── example-icon.svg
└── README.md.liquid
```

## Test plan
- [ ] Verify `shopify app generate extension --template order_attribution_config` scaffolds correctly
- [ ] Verify TOML template renders with correct handle substitution
- [ ] Verify templates.json is valid JSON

Closes https://github.com/shop/issues-multi-channel-management/issues/3126

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Catherine-Fu:
* **#393** <a href="https://app.graphite.com/github/pr/Shopify/extensions-templates/393?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/extensions-templates/393?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

templates.json
```diff
@@ -2068,6 +2068,23 @@
       }
     ]
   },
+  {
+    "identifier": "order_attribution_config",
+    "name": "Order Attribution Config",
+    "defaultName": "order-attribution-config",
+    "group": "Sales channels",
```
jsmarr, line 2075: does this have to be us? :) eventually i hope this isn't channels-related at all...

jsmarr (approved):
I don't think we want this publicly available yet, but not the end of the world if there's no way around that?

templates.json
Catherine-Fu: The group name is shown in the CLI when selecting extension options from different extension groups. It doesn’t have to be “Sales channels”, we can create a new group for the attribution extension. Do you have any suggestions for the group name?

jsmarr: What groups already exist? Anything related to Orders? If it's not clear what to do, we can leave it like this for now and presumably change it later?

Catherine-Fu: here are the existing group options for extensions templates and no order related groups. So will use the "sales channel" for now.

| Group | Count | Description |
| --- | --- | --- |
| **Editor extension collection** | 1 | Editor extensions |
| **Flow** | 4 | Flow automation |
| **Functions** | 13 | Shopify Functions |
| **Links** | 4 | Various link types |
| **Payments extensions** | 6 | Payment extensions |
| **Sales channels** | 2 | Currently only `channel_config` + `order_attribution_config` |
| **Tax calculation** | 1 | Tax calculation |
| **Theme app extension** | 1 | Theme extensions |
| **UI extensions** | 44 | UI extensions |
| **Web pixel extensions** | 1 | Pixel tracking |

jsmarr: Meh let's stick with Sales Channels for now but seems like we could make our own Orders category later if we want to?

Catherine-Fu: > seems like we could make our own Orders category later if we want to?

Yes, a quick investigation shows we only need to set **`group`** field to a new name we prefer in this file.  There's no need to register or declare the group anywhere else. It's free text, and the CLI automatically groups extensions based on the values in templates.json.

I was thinking of group names like “Order configuration” or “Order extensions.”.   Let me know if you have any suggested group names, and I'll update it.

Catherine-Fu:
We [added](https://github.com/shop/world/blob/2db4c482e958815c91f2ad806ecbc020ce3d7684/areas/core/shopify/components/channels/order_attribution/config/initializers/extensions.rb#L20) a beta flag (f_order_attribution_config_extension) to this new extension. So it won't be visible publicly without enabling the beta flag.

PR #394, opened 2026-04-20
Update function extension templates to API version 2026-04

saga-dasgupta:
## Summary
- Bumps `api_version` from `2026-01` to `2026-04` in all 30 Functions extension templates (JS / RS / Wasm variants across cart-checkout-validation, cart-transform, delivery-customization, discount, fulfillment-constraints, location-rules, order-discounts, payment-customization, product-discounts, shipping-discounts).
- Templates pinned to `unstable` (discounts-allocator, pickup-point-delivery-option-generators, local-pickup-delivery-option-generators) are intentionally untouched for the version bump, but their `schema.graphql` files are refreshed alongside everything else.
- Regenerates `schema.graphql` for every function template against the new API version.
- Adds `prerequisites: None` / `"prerequisites": null` to the one Rust source file and one test fixture that construct `ProductDiscountCandidate` literals — the 2026-04 schema adds a nullable `prerequisites: [ProductDiscountCandidatePrerequisite!]` field, and Rust requires every struct field at init even when optional. `cargo check --workspace` is clean after the fix.

The helper script used to regenerate these schemas is in its own PR: #397.

## CI fix: stub `shopify.app.toml` in `.github/scripts/run-wasm-tests.sh`
The Wasm test job installs `@shopify/cli@latest` and generates a minimal `shopify.app.toml` so the CLI can discover extension directories for `buildFunction`. Since PR #383 landed in March, the CLI's app-config validator tightened and now rejects the previous two-field stub with:

```
[client_id]: Required
[auth]: Required
[webhooks]: Required
[application_url]: Valid URL is required
[embedded]: Boolean is required
```

The script now emits a full valid app config (dummy `client_id`, `application_url`, `embedded`, `[auth]`, `[webhooks]`) so validation passes. The values are placeholders — nothing is deployed; the toml only exists for extension discovery.

## Test plan
- [x] `cargo check --workspace` passes after `yarn expand-liquid rust`
- [x] CI passes (JS / TS / Rust / Wasm workflows)
- [x] Spot-check a generated extension compiles and tests pass against 2026-04

### DIFF
--- a/.github/scripts/run-wasm-tests.sh
+++ b/.github/scripts/run-wasm-tests.sh
@@ -22,8 +22,20 @@ esac
 
 # Create a minimal app toml so Shopify CLI can discover the extensions
 cat > shopify.app.toml << EOF
-scopes = ""
+client_id = "00000000000000000000000000000000"
+name = "extensions-templates-ci"
+application_url = "https://example.com"
+embedded = true
 extension_directories = ["${EXT_DIR_PATTERN}"]
+
+[access_scopes]
+scopes = ""
+
+[auth]
+redirect_urls = ["https://example.com/auth/callback"]
+
+[webhooks]
+api_version = "2026-04"
 EOF
 
 for dir in $EXT_DIR_PATTERN; do
--- a/functions-cart-checkout-validation-js/schema.graphql
+++ b/functions-cart-checkout-validation-js/schema.graphql
@@ -120,6 +120,11 @@ type Cart implements HasMetafields {
   ): Attribute
 
   """
+  The billing address associated with the cart.
+  """
+  billingAddress: MailingAddress
+
+  """
   Information about the customer that's interacting with the cart. It includes details such as the
   customer's email and phone number, and the total amount of money the customer has spent in the store.
   This information helps personalize the checkout experience and ensures that accurate pricing and delivery options
@@ -152,6 +157,11 @@ type Cart implements HasMetafields {
   instead.
   """
   deliveryGroups: [CartDeliveryGroup!]!
+
+  """
+  The discounts that have been applied to the cart.
+  """
+  discountApplications: [DiscountApplication!]!
 
   """
   The items in a cart that the customer intends to purchase. A cart line is an entry in the
@@ -193,6 +203,12 @@ type Cart implements HasMetafields {
     """
     namespace: String
   ): Metafield
+
+  """
+  A purchase order number associated with the cart, often used for B2B transactions
+  to reference the buyer's internal purchase order.
+  """
+  poNumber: String
 
   """
   The physical location where a retail order is created or completed.
@@ -250,6 +266,11 @@ type CartDeliveryGroup {
   can choose to have their orders shipped. Examples include express shipping or standard shipping.
   """
   deliveryOptions: [CartDeliveryOption!]!
+
+  """
+  The discounts that have been applied to the delivery group.
+  """
+  discountAllocations: [DiscountAllocation!]!
 
   """
   The type of merchandise in the delivery group.
@@ -358,6 +379,11 @@ type CartLine {
   the same t-shirt to their cart, then each size is represented as a separate cart line.
   """
   cost: CartLineCost!
+
+  """
+  The discounts that have been applied to the cart line.
+  """
+  discountAllocations: [DiscountAllocation!]!
 
   """
   The ID of the cart line.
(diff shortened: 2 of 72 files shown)


lopert (approved):

refresh-function-schemas.sh
```diff
@@ -0,0 +1,134 @@
...
+  if ! only_match "$ID"; then
+    skipped=$((skipped + 1)); continue
+  fi
+
+  EXT="schemagen-${ID//_/-}"
+  rm -rf "extensions/$EXT"
+  cleanup_list+=("$EXT")
+
+  echo ">> $ID — generating from $BRANCH"
+  shopify app generate extension --template "$ID" --flavor rust -n "$EXT" -u "$CLONE_URL"
```
adampetro: why do we need to generate from github if we have everything here locally already?

saga-dasgupta: Because I made it so that if you can make it use any branch to generate templates, including ones you don't have checked out locally. Maybe that is an unnecessary feature that I can remove and make it so it only looks at the current local working tree.

saga-dasgupta: Removed it.

saga-dasgupta: I was anyway on edge about adding it but thought maybe it could be useful at some point so I am happy to remove it.

achitojha (approved):

refresh-function-schemas.sh
adampetro: Yeah I don't think it is necessary, if we need to run against a specific branch we can check it out

refresh-function-schemas.sh
```diff
@@ -0,0 +1,133 @@
...
+
+refreshed=0
+skipped=0
+for row in "${ROWS[@]}"; do
+  ID="$(printf '%s' "$row" | cut -f1)"
+  FIRST_PATH="$(printf '%s' "$row" | cut -f2)"
+  [ -n "$ID" ] && [ -n "$FIRST_PATH" ] || continue
+
+  PREFIX="${FIRST_PATH%-*}"
+  RS_DIR="$TEMPLATES_REPO/${PREFIX}-rs"
```
adampetro: why do we need to do anything language specific here? we shouldn't care about language when updating the schema

refresh-function-schemas.sh
```diff
@@ -0,0 +1,133 @@
...
+  if [ ! -f "$RS_DIR/shopify.extension.toml" ]; then
+    echo "SKIP $ID (no expanded toml at $RS_DIR)"
+    skipped=$((skipped + 1)); continue
+  fi
+
+  EXT="schemagen-${ID//_/-}"
+  LINK="$APP_DIR/extensions/$EXT"
+  rm -rf "$LINK"
+  ln -s "$RS_DIR" "$LINK"
+  cleanup_list+=("$EXT")
```
adampetro: Why do we need to do this in a separate directory with symlinks instead of just doing it inline?

refresh-function-schemas.sh
adampetro, line 1: Let's start with something simple and easy to read. We can always add more options and customizability later if there is a need

adampetro (changes requested):

refresh-function-schemas.sh
saga-dasgupta: The templates expansion step still has to pick some flavor to run against, I just picked Rust because it has the simplest expansion path, and then since the schema is the same regardless of language I copy it back for all the different language flavors.

refresh-function-schemas.sh
saga-dasgupta: Because `shopify app function schema` walks up from its cwd looking for a `shopify.app.toml`, then uses that linked app's `client_id` to make an authenticated call to the Partners API and fetch the schema for the given api_version. So it has to be run from inside a directory that sits under a real linked app, not from this templates repo. The symlink is the cheap way to satisfy that constraint: we put a pointer to the template dir under `<linked-app>/extensions/schemagen-*` so the CLI can discover it as a function extension of the app, without copying source files around. When schema writes back to the extension dir, we copy the result into all three flavor folders in the repo and delete the symlink. Running inline would mean either (a) treating this repo as a Shopify app (which it isn't), or (b) authenticating against a dummy app and routing the request, both of which add more machinery than a one-line `ln -s`.

refresh-function-schemas.sh
saga-dasgupta: Sure, what exactly do you want me to remove?

refresh-function-schemas.sh
adampetro: Could we copy the `shopify.app.toml` to the root directory of this repo because [it is `.gitignored`](https://github.com/Shopify/extensions-templates/blob/c09702548b3e16550119b43170520799a84aafa2/.gitignore#L80)? Or do we consider that too potentially risky?

refresh-function-schemas.sh
adampetro: I realize that it is wasteful in the event that all flavors have the same API version, but we don't have any constraint enforcing that so I think it is most correct to generate for each flavor (it's also possible we add more nuance to the `function generate schema` command based on the extension toml so I think it is safest to always generate the schema per-template)

saga-dasgupta: Oh, I didn't think there would be a case where we have different API versions for different languages of the same function? I suppose we could but how/why would that be? I am not disagreeing, I can make it the way you suggested, I was just trying to make it the most efficient for what I thought would almost always be the case.

adampetro: I don't think we would ever do it intentionally, but I don't like making assumptions that aren't enforced. If we want to make that assumption then I think we should add some validation to ensure that is the case

saga-dasgupta: Okay, yeah thats fair.

refresh-function-schemas.sh
saga-dasgupta: It's actually what `run-wasm-tests.sh` does already, writing a stub `shopify.app.toml` at the root for test discovery. The issue is that `shopify app function schema` hits the Partners API, so it needs a real `client_id` a developer has org access to. That means each dev would do shopify app config link once inside this repo to drop their own gitignored `shopify.app.toml` which I think is a lot more friction than what I have in place right now. And then there's also the issue that the repo root becomes a linked Shopify app. Running something like `shopify app deploy` from the root would try to deploy every function template as a real extension. I like the solution I have in place more.

refresh-function-schemas.sh
saga-dasgupta: This is done, now they all get expanded individually to generate their own schemas.

adampetro (approved):

adampetro:
Can we update the description based on the removal of the script?

saga-dasgupta:
> Can we update the description based on the removal of the script?

It already is updated.

adampetro:
> It already is updated.

my bad, mixed up the `shopify.app.toml` comment with the schema update script but it is actually the fix for the existing script for running the wasm tests

PR #406, opened 2026-05-14
remove static assets from template for backwards compatibility

JoshuaWhite1:
### Background

Closes https://github.com/shop/issues-admin-extensibility/issues/2546

https://github.com/Shopify/shopify-app-template-extension-only/pull/29

Related: 
### Solution

### Checklist


melissaluu (approved):
Code looks good

PR #409, opened 2026-05-20
Ensure we await promises when forms are submitted

JoviDeCroock:
- This ensures that form submission is awaited with `event.waitUntil`
- We ensure that updates by other users/modals/... can't overwrite state by loading the faq's before re-submitting


melissaluu (approved):
🎩 & code looks good

PR #412, opened 2026-05-25
Move App home template after Checkout UI

davejcameron:
### Background

Moving an item in the list from bottom to near the top

### Solution

### Checklist

### DIFF
--- a/templates.json
+++ b/templates.json
@@ -116,7 +116,7 @@
     "organizationExpFlags": [
       "d867a776"
     ],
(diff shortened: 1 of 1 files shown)


JoshuaWhite1 (approved):
Should we maybe group it with action/block extensions around 384 to keep it near our other stuff?  Either way LG2M!

melissaluu (approved):
Code changes looks good but I haven't been able to verify it locally.

melissaluu (approved):

PR #418, opened 2026-07-13
Add routing to app-home extension template

MitchLillie:
TL;DR: This adds the same URL-based routing capability to the generated App Home extension template that the extension-only app template already demonstrates.

## What changed

- Adds `preact-iso` to the `app-home` template dependencies.
- Wraps the template in `LocationProvider`, `ErrorBoundary`, and `Router`.
- Splits the sample into route components:
  - `/` for the FAQ list
  - `/faq/new` for the FAQ creation form
  - `/faq/:id` for editing an existing FAQ
  - `/settings` as an `s-app-nav` sub-navigation example
  - default not-found route
- Keeps the Storage API sample, but moves add/list/edit navigation from local `useState` view switching to router navigation.
- Adds edit affordances from the list view and an edit form route for existing FAQs.
- Splits the generated source into multiple files so the route structure is easier to follow:
  - `src/AppHome.jsx`
  - `src/pages/HomePage.jsx`
  - `src/pages/FAQPage.jsx`
  - `src/pages/SettingsPage.jsx`
  - `src/pages/NotFoundPage.jsx`
  - `src/models/faqs.js`
  - `src/components/AboutTemplate.jsx`
- Updates README copy to describe URL-based routing, `s-app-nav`, and the generated file structure.

## Why this shape

The extension-only app template already ships a routed App Home. This brings the `shopify app generate extension` path closer to that capability while keeping the generated extension self-contained: no shared app scaffold, no app-level metaobject setup, and no API routes required.

## Validation

- `shopify app init --template remix --flavor javascript --package-manager pnpm`
- `shopify app generate extension --template app_home --name app-home --clone-url https://github.com/Shopify/extensions-templates#mlillie/app-home-router-template`
- Confirm the generated source tree includes the split route files listed above.
- `shopify app dev`
- Confirm that all FAQ functions work: create, delete, edit.

### DIFF
--- a/app-home/README.md.liquid
+++ b/app-home/README.md.liquid
@@ -2,7 +2,8 @@
 
 A simple app home extension that demonstrates:
 - Basic UI with Shopify's Remote DOM components
-- Simple client-side routing (list view / add view)
+- URL-based client-side routing with preact-iso
+- App sub-navigation with `s-app-nav`
 - KV storage for persisting data
 
 ## Features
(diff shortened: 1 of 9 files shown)


melissaluu (approved):
🎩 and code looks good!

PR #419, opened 2026-07-15
Add Discounts app action link extension template

loic-d:
## Problem

There is no template for partners who want to expose a **discount** app action to Sidekick. The only app action link template today is the `app_action_link` (email) one.

## What this PR does

Adds a new `discount_app_action_link` template (folder `discount-app-action-link`) that registers a `shopify/Discount` `create` app action link intent for Sidekick, modeled on the existing `app_action_link` (email) template. When a merchant asks Sidekick to create a discount an app powers, Sidekick navigates the merchant to the right page in that app to complete it.

The template ships behind the `discount_app_action_link` organization beta flag, so it stays hidden until the partner org has the beta enabled.

New files:
- `discount-app-action-link/` — README, instructions, intent schema, tools, locales, and `shopify.extension.toml.liquid`
- `templates.json` — registers the template under the Sidekick group

## What's next

Enable the `discount_app_action_link` beta flag for partner orgs once ready.

## Test plan

- `templates.json`, `intent-schema.json`, and `tools.json` validated as well-formed JSON.
- Template folder structure mirrors the existing `app_action_link` template.


iphipps (approved):
should we also add a template for the edit extension?

loic-d:
@iphipps good call — added a dedicated `discount_app_action_link_edit` template (folder `discount-app-action-link-edit`) in 3513d83.

Kept it a separate template rather than a flavor of the create one, since each `admin.app.intent.link` extension registers exactly one intent and an app may want both create and edit — so they need distinct handles. This mirrors the DFT reference (`dft-discount-intent-link` + `dft-discount-intent-link-edit`).

The edit intent is navigation-only: it maps the discount GID to the `:id` route param via a top-level `value` block and derives the function from the discount, so there's no `functionId` `matchValue` and no `tools.json`/`instructions.md`. Both templates share the `discount_app_action_link` org beta flag. I also renamed the create display name to "Discount app action link (create)" for symmetry and pointed the create README's edit section at the new template.

templates.json
```diff
@@ -2137,6 +2137,48 @@
...
+    "extensionPoints": [],
+    "supportedFlavors": [
+      {
+        "name": "Config only",
+        "value": "config-only",
+        "path": "discount-app-action-link"
+      }
+    ],
+    "organizationExpFlags": [
+      "a9b1bd3a"
```
iphipps, line 2157: 👍

iphipps (approved):

discount-app-action-link/tools.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "name": "create_discount",
+    "description": "Create a discount powered by this app's discount function.",
+    "inputSchema": {
+      "type": "object",
+      "properties": {},
```
iphipps: 👍 
ref: https://github.com/Shopify/discount-functions-testing/pull/1398

discount-app-action-link-edit/README.md
```diff
@@ -0,0 +1,48 @@
+# Sidekick Discount App Action Link Extension (Edit)
+
+Sidekick app action link extensions let app developers expose actions to Sidekick in the Shopify Admin. This template registers an **edit** discount app action (an `edit` intent for the `shopify/discount` type) so that when a merchant asks Sidekick to change a discount your app powers, Sidekick navigates the merchant to the right page in your app to edit it.
+
+This is the edit counterpart to the [`discount_app_action_link`](../discount-app-action-link) (create) template. Each `admin.app.intent.link` extension registers exactly one intent, so create and edit are separate extensions with separate handles. Generate whichever intents your app supports.
```
vividviolet: They need to support both otherwise the full discount creation and edit flow will not work. We should combine both edit and create into the same extension folder and have the TOML contain 2 extensions instead. This way when generating the extension from a template they have a good starting point with 2 intents.

discount-app-action-link/tools.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "name": "create_discount",
```
vividviolet: Tools are not supposed to repeat the intent. Check out the other templates for a better example of a tool. They are supposed to let sidekick work with the app to make client side changes.

discount-app-action-link-edit/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+### How it works
+
+1. The extension registers an intent via `[[extensions.targeting.intents]]` with `type = "shopify/discount"` and `action = "edit"`.
+2. `intent-schema.json` uses the `shopify-intent.json` meta-schema. A top-level `value` block maps the discount GID to the `id` route param (`fieldName: "id"`), so Sidekick fills the `:id` in the URL. The `outputSchema` returns the edited discount's GID.
+3. When Sidekick invokes the action, the merchant is navigated to your app at the configured `url` with the discount id filled in.
+
+### Why no `functionId`
+
+Unlike the create template, the edit route derives the function from the existing discount, so there's no `functionId` `matchValue` to pin. The intent only needs the discount id. That also means the edit intent is navigation-only — it doesn't ship a `tools.json` or `instructions.md`.
```
vividviolet: Why isn't there a functionId? If we don't pass it then the app would need to do the look up itself which seems like additional work. We should make it match the create flow and pass `functionId` as an input.

discount-app-action-link/tools.json
```diff
@@ -0,0 +1,20 @@
+[
```
lauramann: As part of https://app.graphite.com/github/pr/shop/world/876126 we will be removing the requirement for `tools`. Instead of adding a dummy tool here we should be able to remove this entirely. cc @bahar-p

loic-d:
Reopened as a same-repo PR (was from a fork): #420 — same content, squashed to one commit, now with `shopify/Discount` (PascalCase) casing. Please continue review on #420. 🙏

PR #422, opened 2026-07-29
Bump checkout and accounts extensions to 2026-07

andrewmcgov:
### Background

Update checkout and accounts extension templates to 2026.7.x


robin-drexler (approved):
good call :D

PR #424, opened 2026-08-28
Update extension templates to Bundler module resolution

akhayoon:
## Background

Resolves shop/issues-retail#33804.

Generated POS Preact extensions import `@shopify/ui-extensions/preact`, but their legacy Node10 module resolution does not understand that package's modern `exports` entry. TypeScript 6 also rejects the deprecated `node` resolution mode with `TS5107`.

## Solution

Replace `"moduleResolution": "node"` with `"moduleResolution": "Bundler"` in every generated extension configuration that still uses the legacy resolver.

This is intentionally a resolver-only change:

- All 18 affected templates keep their existing `target` and `noEmit` settings, and continue to omit `module`.
- No explicit module policy is introduced by this PR.
- The existing discount settings template remains unchanged on `ESNext` plus `Bundler`.
- TypeScript 5 or newer is required because `Bundler` was introduced in TypeScript 5.

Existing generated projects are not modified.

The exact tested head is `d5352b09dc5f4a05cb053be7812f1dd4c9862771`.

## Testing

Verified from exact base `c05f026ad348e1c340b06effe6ba7a51ee0cffab` against the final candidate:

- The final diff is exactly 18 `moduleResolution` replacements from `node` to `Bundler`, with no other changed lines.
- Shopify CLI 4.7.1 rendered all 134 registry selections represented by 74 unique path/flavor pairs twice. Each side produced 550 deterministic files.
- The 17 registry-reachable changed configurations plus the unregistered Preact Liquid branch retain `target: ES2020`, omit `module`, use `moduleResolution: Bundler`, and retain `noEmit: true`.
- TypeScript 5.0.4, 5.9.3, and 6.0.3 accept the options in every migrated configuration. Current-head and resolver-only candidate source diagnostics are identical for all 17 reachable configurations under all three compilers.
- For an exact generated POS Preact extension, TypeScript 5.9.3 fails to resolve `@shopify/ui-extensions/preact` with Node10 and resolves it to the package's exported `build/ts/preact.d.ts` entry with Bundler.
- TypeScript 6.0.3 reports `TS5107` for the base POS configuration and reports no diagnostics for the resolver-only candidate.
- Twenty real `shopify app build` commands covered all 70 valid generated cases twice: ten current-head controls and ten resolver-only candidate builds. All 75 JavaScript files, 70 source maps, 75 metafiles, and 18 generated declarations were byte-identical between control and candidate and deterministic across repeat builds.
- `git diff --check` passes. No legacy `moduleResolution: "node"` setting remains.

TypeScript 4.9 and older cannot read `moduleResolution: "Bundler"`; TypeScript 5 or newer is the intentional baseline for newly generated extensions.

This PR deliberately leaves `module` omitted as it was before. TypeScript 5 derives ES2015 from the existing ES2020 target, while TypeScript 6 derives ES2020. Defining a version-independent module syntax policy is separate from this resolver ticket. The affected template sources currently use neither `import.meta` nor dynamic imports.

## Checklist

_Co-authored by AI._


akhayoon:
### A/B verification

I rendered the Preact `pos-action` template, installed its real declared dependencies (`@shopify/ui-extensions@2025.10.16`, Preact, and Signals), and tested with TypeScript 6.0.2.

**Before — current `main` template:**

```text
tsconfig.node.json(8,25): error TS5107: Option `moduleResolution=node10` is deprecated and will stop functioning in TypeScript 7.0.
```

**After — this PR:**

The focused import-resolution check exits successfully, and TypeScript reports:

```text
Module name `@shopify/ui-extensions/preact` was successfully resolved to `.../@shopify/ui-extensions/build/ts/preact.d.ts` with Package ID `@shopify/ui-extensions/build/ts/preact.d.ts@2025.10.16`.
```

This directly verifies the behavior changed by this PR: generated POS templates use a supported resolver and TypeScript can locate the required Preact entry point.

henryStelle:
It looks like there are lots more templates with this issue than just ours from [https://github.com/search?q=repo%3AShopify%2Fextensions-templates+"moduleResolution"%3A+"node"&type=code](https://github.com/search?q=repo%3AShopify%2Fextensions-templates+%22moduleResolution%22%3A+%22node%22&type=code). Should we make this a wider update? And if so, use that to align on these being the correct config options (that is what the AI recommended, but I haven't dug into if its right)

akhayoon:
Great callout @henryStelle. I added for the rest.

henryStelle (approved):
I looked up these tsconfig options and these look like the correct changes. It would probably be good to get an 👍 from at least one of the other extension teams.

js-goupil (approved):

akhayoon:
### Exact-head terminal verification

Read-only local reproduction against `64c651898afd5d278714661c85db0f91531f57b3`.

- Final scope: 18 intended migrations, discount template absent.
- Shopify CLI: 20/20 build commands passed; all compared artifacts are byte-identical.
- TypeScript: migrated configs accepted by 5.0.4, 5.9.3, and 6.0.3; the `import.meta` witness is rejected with `TS1343`.

These are local reproduction results, not hosted CI results.

Exact final head, changed-file scope, and live PR state

Shopify CLI build commands and byte-for-byte artifact comparison

TypeScript configuration matrix and import.meta rejection

akhayoon:
### Exact-head resolver-only verification

Local read-only reproduction against `d5352b09dc5f4a05cb053be7812f1dd4c9862771` after limiting PR #424 to the ticket's resolver change.

- Scope: exactly 18 `moduleResolution` replacements from `node` to `Bundler`; no explicit `module` additions.
- Ticket witness: TypeScript 6.0.3 reports `TS5107` for the generated POS base configuration and no diagnostics for the candidate. TypeScript 5.9.3 changes `@shopify/ui-extensions/preact` from unresolved under Node10 to its exported declaration under Bundler.
- Rendering: 134 registry selections, 74 unique cases, 550 deterministic files per side, with only the expected 17 reachable configurations plus one unregistered Liquid branch changed.
- Shopify CLI: 20/20 control and candidate build commands passed across 70 valid cases. All 238 compared artifacts per run were byte-identical and repeatable.
- TypeScript: 5.0.4, 5.9.3, and 6.0.3 accept every migrated configuration's options; current-head and candidate source diagnostics match exactly.

These are local reproduction results, not hosted CI or live deployment results.

Exact head and resolver-only scope

Shopify CLI build matrix

TypeScript resolver proof

